PR #8, opened 2017-12-15
increase the font size of admin UI

abbyck:


abbyck:
@nemesisdesign Can you please check this.

abbyck:
@nemesisdesign I thought my only duty is to change the font. I too had that idea of adjusting padding. I will try as you said.
@rohithasrk point no:1 , by tabs you mean the tables, right?

nemesisdesign:
@abbyck you are right in the sense that the text of the tasks indicates only font-size, although every detail in a layout cannot be changed alone without tweaking also related details. Anyway I hope you can improve your patch so we can use it ;-)

Issue #6, opened 2017-12-14
Improve instructions for installation during development


rohithasrk:

abbyck:
I would like to improve that

nemesisdesign:
@rohithasrk could we add some detail on what needs to be done? Shall we turn this into a GCI task?

rohithasrk:
@nemesisdesign Yeah. Can be a GCI task

nemesisdesign:
We should replace the line `python setup.py develop` with `pip install -e .[users]`, which will automatically install openwisp-users, which depends on django and will install the right version of django.

nemesisdesign:
Added a task for this

abbyck:
Oops. I missed your reply @nemesisdesign. If I recall correctly I deleted a comment that I made in between that I will fix this issue. Sorry for the mention :)

nemesisdesign:
lots of GCI students have been able to contribute to this module in the last 2 years, I think the README now is good enough so I'm closing this issue

PR #11, opened 2018-01-07
Improved installation instructions

UtkarshVerma:
Implemented the explicit usage of `python3` and also added a few module installations in `requirements-test.txt`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -225,7 +225,7 @@ Add ``openwisp_utils.staticfiles.DependencyFinder`` to ``TEMPLATES_LOADERS`` in 
 Installing for development
 --------------------------
 
-To install openwisp-utils, you must have python3 installed.
+To install openwisp-utils, you must have python3 installed, since Django 2.0 is going to be used and it isn't supported by python2.7.
 
 Install sqlite:
 
--- a/README.rst
+++ b/README.rst
@@ -73,7 +73,7 @@ If you want to contribute, install your cloned fork:
 
     git clone git@github.com:<your_fork>/openwisp-utils.git
     cd openwisp-utils
-    python setup.py develop
+    python3 setup.py develop
 
 Using the utilities in OpenWISP modules
 ---------------------------------------
@@ -225,6 +225,8 @@ Add ``openwisp_utils.staticfiles.DependencyFinder`` to ``TEMPLATES_LOADERS`` in 
 Installing for development
 --------------------------
 
+To install openwisp-utils, you must have python3 installed.
+
 Install sqlite:
 
 .. code-block:: shell
@@ -237,21 +239,21 @@ Install your forked repo:
 
     git clone git://github.com/<your_fork>/openwisp-utils
     cd openwisp-utils/
-    python setup.py develop
+    python3 setup.py develop
 
 Install test requirements:
 
 .. code-block:: shell
 
-    pip install -r requirements-test.txt
+    pip3 install -r requirements-test.txt
 
 Create database:
 
 .. code-block:: shell
 
     cd tests/
-    ./manage.py migrate
-    ./manage.py createsuperuser
+    python3 manage.py migrate
+    python3 manage.py createsuperuser
 
 Set ``EMAIL_PORT`` in ``settings.py`` to a port number (eg: ``1025``):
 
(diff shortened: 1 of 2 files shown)


UtkarshVerma:
Solves the issue #6

h0n3yc4k3:
I think that you should left *python*, instead of *python3*
Because I think that you should use virtualenv (from my experience)

UtkarshVerma:
Yes, that is how I do it. But instructions are always supposed to be written in a basic and clear manner. Here python3 reminds the user that python3 is being used. It's up to him to use virtualenv or not

UtkarshVerma:
Yes, that is how I do it. But instructions are always supposed to be written in a basic and clear manner. Here python3 reminds the user that python3 is being used. It's up to him to use virtualenv or not

README.rst
```diff
@@ -73,7 +73,7 @@ If you want to contribute, install your cloned fork:
 
     git clone git@github.com:<your_fork>/openwisp-utils.git
```
h0n3yc4k3: I think *<your_username>* will be better, than *<your_fork>* but it is nit-picking :P

UtkarshVerma:
Yes, that is how I do it. But instructions are always supposed to be written in a basic and clear manner. Here python3 reminds the user that python3 is being used. It's up to him to use virtualenv or not

UtkarshVerma:
Yes, that is how I do it. But instructions are always supposed to be written in a basic and clear manner. Here python3 reminds the user that python3 is being used. It's up to him to use virtualenv or not

README.rst
```diff
@@ -263,17 +265,17 @@ Launch development server and SMTP deubgging server:
...
+    python3 -m smtpd -n -c DebuggingServer localhost:1025
 
 You can access the admin interface of the test project at http://127.0.0.1:8000/admin/.
 
 Run tests with:
 
 .. code-block:: shell
 
-    ./runtests.py
+    python3 runtests.py
```
h0n3yc4k3: If you will use virtualenv then you dont need to replace it :)

h0n3yc4k3:
But you can just specify which version of the python is working correctly

README.rst
UtkarshVerma: I know that. But still thanks.

README.rst
UtkarshVerma: Nitpicking is also necessary.

h0n3yc4k3:
And I know that @nemesisdesign like when you squash your commits and choose name of guidelines like *[docs] Updated README* :)

UtkarshVerma:
Fixes #6

README.rst
```diff
@@ -237,21 +239,21 @@ Install your forked repo:
 
     git clone git://github.com/<your_fork>/openwisp-utils
     cd openwisp-utils/
-    python setup.py develop
```
h0n3yc4k3: @nemesisdesign said: we should replace the line `python setup.py develop` with `pip install -e .[users]`, which will automatically install openwisp-users, which depends on django and will install the right version of django
Did you take this opinion into account?

UtkarshVerma: No

README.rst
nemesisdesign: don't do this please

README.rst
nemesisdesign: @h0n3yc4k3 is right

UtkarshVerma: ok, will test it on a new virtualenv and make the necessary commits

README.rst
UtkarshVerma: Ok, but just for reference purposes, why?

h0n3yc4k3: Maybe he too should just introduce for virtualenv ?

UtkarshVerma: Yeah, I was thinking about that

README.rst
UtkarshVerma: @nemesisdesign @h0n3yc4k3 the `pip install -e .[users]` fetched Django 2.0 as well

h0n3yc4k3: Yes, default installing on python2 produce error **AttributeError: 'module' object has no attribute 'lru_cache'** because it try to install django2 by default (which not supported).

But you can add note like "If you are using python2 then you shoud run `pip install django~=1.0` (or any 2>django>=1.8) before `pip install -e .[users]`" (for bypassing installing django2)
I didnt check comptability, but it installed :D I'll check it

PR #13, opened 2018-08-19
BaseSerializer Implemented

### LINKED ISSUE #12: Implement a base serializer class for API views
Django-rest-framework, which has been used to create API views, does not call the `full_clean` method under models, this means that custom model validations would not run on API views while creating or updating records. We need to create a base serializer class which can be extended to other apps.

atb00ker:
Added new file `openwisp_utils/api/serializers.py` and implemented a base serializer that can be used in other projects.

Fixes #12

### DIFF
--- a/openwisp_utils/api/serializers.py
+++ b/openwisp_utils/api/serializers.py
@@ -1,0 +1,14 @@
+from django.core.exceptions import ImproperlyConfigured
+
+try:
+    from rest_framework import serializers
+except ImportError:  # pragma: nocover
+    raise ImproperlyConfigured('Django REST Framework is required to use '
+                               'this feature but it is not installed')
+
+
+class ValidatedModelSerializer(serializers.ModelSerializer):
+    def validate(self, data):
+        instance = self.instance or self.Meta.model(**data)
+        instance.full_clean()
+        return data
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -4,3 +4,4 @@ isort
 flake8
 # For testing Dependency loaders
 django_netjsonconfig
+djangorestframework>=3.8.2,<3.9
--- a/tests/test_project/models.py
+++ b/tests/test_project/models.py
@@ -1,3 +1,4 @@
+from django.core.exceptions import ValidationError
 from django.db import models
 from django.utils.translation import ugettext_lazy as _
 from openwisp_users.mixins import OrgMixin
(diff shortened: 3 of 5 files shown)


setup.py
```diff
@@ -33,7 +33,7 @@
     packages=find_packages(exclude=['tests', 'docs']),
     include_package_data=True,
     zip_safe=False,
-    install_requires=['django-model-utils'],
+    install_requires=['django-model-utils', 'djangorestframework'],
```
nemesisdesign: not all the packages using `openwisp-utils` have APIs and adding this here would make them download DRF even when not needed.

I suggest the following solution:

- remove this requirement here
- add djangorestframework to `requirements-test.txt`
- add a try/except `ImportError` block when you import `serializers` in `openwisp_utils/api/serializers.py`, see my other comment

openwisp_utils/api/serializers.py
```diff
@@ -0,0 +1,8 @@
+from rest_framework import serializers
```
nemesisdesign: following the text of the other comment, here you could do something like:

```python
try:
    from rest_framework import serializers
except ImportError:  # pragma: nocover
    import sys
    print('Django REST Framework is required to use this feature but it is not installed')
    sys.exit(1)
```

atb00ker: How about using django's abstraction for this?

```
from django.core.exceptions import ImproperlyConfigured

try:
    from rest_framework import serializers
except ImportError:  # pragma: nocover
    raise ImproperlyConfigured('Django REST Framework is required to use this feature but it is not installed')
```

nemesisdesign: sure, a lot better! :+1:

atb00ker:
Made the required changes. :smile:

openwisp_utils/api/serializers.py
```diff
@@ -0,0 +1,14 @@
+from django.core.exceptions import ImproperlyConfigured
+
+try:
+    from rest_framework import serializers
+except ImportError:  # pragma: nocover
+    raise ImproperlyConfigured(
+        'Django REST Framework is required to use this feature but it is not installed')
```
nemesisdesign: we usually break this kind of code into multiple lines with something like:

```python
raise ImproperlyConfigured('Django REST Framework is required to use '
                           'this feature but it is not installed')
```

Could you update these 2 lines for consistency reasons please?

PR #16, opened 2018-09-30
[qa] Added check migration filename script

### LINKED ISSUE #15: [enhancement] adding ci test for checking migration name
Add ci test that checks if migration file created has a descriptive name, fail test if default name pattern is found.

atb00ker:
- [x] Implement function to check migration file name is not in default pattern
- [x] Write Tests

Closes #15

### DIFF
--- a/openwisp_utils/ci.py
+++ b/openwisp_utils/ci.py
@@ -1,0 +1,71 @@
+"""
+This file contains functions that are required by
+multiple openwisp repositories for testing
+during continuous integration
+"""
+import argparse
+import os
+import re
+
+
+def _parse_args():
+    """
+    Parse and return CLI arguements
+    """
+    parser = argparse.ArgumentParser(description="Ensures migration files "
+                                     "created have a descriptive name. If "
+                                     "default name pattern is found, "
+                                     "raise exception!")
+    parser.add_argument('--no-migration-name', nargs='?', const=True,
+                        help="Skip check: migration files have a "
+                        "descriptive name.")
+    parser.add_argument("--migration-path",
+                        help="Path to `migrations/` folder")
+    parser.add_argument("--migrations-to-ignore",
+                        type=int,
+                        help="Number of migrations after which checking of "
+                        "migration file names should begin, say, if checking "
+                        "needs to start after `0003_auto_20150410_3242.py` "
+                        "value will be `3`")
+    return parser.parse_args()
+
+
+def check_migration_name(migration_path, migrations_to_ignore=0):
+    """
+    Ensure migration files created have a descriptive
+    name; if default name pattern is found, raise exception
+    Arguements:
+      - migration_path: path to `migrations/` folder
+      - migrations_to_ignore: number of migrations after
+        which checking should begin, say, if checking needs to
+        start after `0003_auto_20150410_3242.py` value will be `3`
+    """
+    migrations_set = set()
+    migrations = os.listdir(migration_path)
+    for migration in migrations:
+        if (re.match(r"^[0-9]{4}_auto_[0-9]{8}", migration) and
+                int(migration[:4]) > migrations_to_ignore):
+            migrations_set.add(migration)
+
+    if bool(migrations_set):
+        raise Exception("Migration files %s in folder %s need to "
+                        "be renamed to something more descriptive!"
+                        % (str(migrations_set), migration_path))
+
+
+def initialize():
+    """
+    Get & check if CLI arguements are passed
+    correctly and call CI Tasks
+    Intented to be called from setup.py
+    """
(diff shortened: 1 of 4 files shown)


openwisp_utils/ci.py
```diff
@@ -0,0 +1,34 @@
+"""
+This file contains functions that are required by
+multiple openwisp repositories for testing
+during continuous integration
+"""
+import os
+import re
+
+
+def CheckMigrationName():
```
nemesisdesign: keep in mind camel case notation should be used only for classes

openwisp_utils/ci.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+    Check if the migration files created have a descriptive
+    name; if default name pattern is found, exit with 1
+    """
+    MIGRATION_PATH = os.environ["MIGRATION_PATH"]
+    PROJECT_NAME = os.environ["TRAVIS_REPO_SLUG"]
```
nemesisdesign: since this is a function you don't need uppercase variable names (which is a convention to mean constants and global constants), you should use all lowercase variable names

openwisp_utils/ci.py
```diff
@@ -0,0 +1,34 @@
...
+
+    # Get INGORE_SET
+    if "openwisp-users" in PROJECT_NAME:
+        IGNORE_SET = set(["0002_auto_20180508_2017.py"])
```
nemesisdesign: I don't remember how many migrations contain `auto` in the openwisp2 modules, this would become hard to maintain.
I think we can do something simpler which will still help us out: check only the last `<n>` files (in alphabetical order) and raise an issue only if one of that file contains. The `<n>` can be passed to the script as an argument and would have the default of `1`. What do you think?

tests/test_project/tests/test_ci.py
```diff
@@ -0,0 +1,23 @@
...
+        environment = EnvironmentVarGuard()
+        environment.set('MIGRATION_PATH', './tests/test_project/migrations/')
+        environment.set('TRAVIS_REPO_SLUG', 'Test')
+        CheckMigrationName()
+
+    def test_migration_name_fail(self):
+        environment = EnvironmentVarGuard()
+        environment.set('MIGRATION_PATH', './tests/test_project/migrations/')
+        open('./tests/test_project/migrations/0002_auto_20181001_0421.py', 'w').close()
+        CASES = ['test', 'openwisp-users']
```
nemesisdesign: same here regarding uppercase variable names

openwisp_utils/ci.py
atb00ker: Oops, i started this task with a shell script and copied all the uppercase variables from there! 
I am on fixing this! :smile:

atb00ker:
I am testing this for openwisp-users! :smile:

openwisp_utils/ci.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+    Check if the migration files created have a descriptive
+    name; if default name pattern is found, exit with 1
+    Arguements:
+      - migration_path: absolute path to `migrations/` folder
+      - migrations_to_ignore: number of migrations after
+        which checking should begin, say, if checking needs to
+        start after `0003_auto_20150410_3242.py` value will be `3`
+    """
+    migrations_to_ignore = migrations_to_ignore
```
rohithasrk: Why this line @atb00ker

atb00ker: Oops, it was unintentional. Thanks! :blush:

atb00ker:
@nemesisdesign 
> I don't understand this problem of not being able to run checks on our machine.

After reading the [comment about removing code from a repo/package](https://github.com/openwisp/ansible-openwisp2/pull/100#issuecomment-425384578), i was trying to implement the solution provided there for this problem. However, the problem with that approach is that testing will happen only when `.travis.yml` runs and not on the workbench(like `flake8`). That's what i meant.

Keeping this point in mind, i changed that code.
The new idea was running the tests with `runtests.py` as implemented [here](https://github.com/openwisp/openwisp-users/pull/35/files) which would enable us to test the errors on our workbench as well.

> I like the separate script allows to do the checks in separate phases of development

The solution of running it like `flake8` and `isort` sounds much better than adding it in the test suit. it will allow us to run these quality checks separately & making it a standalone python package sounds great.
Thanks for the resources linked. I'll get started with implementing it! :blush:

nemesisdesign:
@atb00ker great

atb00ker:
@nemesisdesign Great! Thanks! :smile: 
I tried it with openwisp-users: https://github.com/openwisp/openwisp-users/pull/35/files
It seems to work! :blush:

PR #19, opened 2018-11-16
[qa] Minor improvements of naming of migration check

nemesisdesign:


nemesisdesign:
@atb00ker check when you can

atb00ker:
One question for `call_check_migration_name()`; actually, i named it `initialize()` because i thought there are more tests that can be called from `initialize()` ( like checking if there are pending migrations -- or any such django related test we might in need in future ).

Hence, with that in mind, i added [`args.no_migration_check`](https://github.com/openwisp/openwisp-utils/pull/19/files#diff-9b88a159c401bd3f6f21d62afa559dc6R70) to give user the option to avoid `check_migration_name`. However, if `call_check_migration_name()` is only planned to call `check_migration_name()`, that option is obsolete. (and maybe that code in a separate function doesn't make sense.)

Should i change it accordingly ?

atb00ker:
Removed option `--no-migration-check`, function `call_migration_name_check()` and test cases rendered obsolete due to changes. :smile:

PR #22, opened 2018-11-21
[WIP] [tests] Increase test coverage to 96%

oltarzewskik:
GCI task: https://codein.withgoogle.com/dashboard/task-instances/4713943676223488/


OltarzewskiK:
Is there any way to test this [line](https://github.com/openwisp/openwisp-utils/blob/master/openwisp_utils/admin.py#L38)

OltarzewskiK:
@atb00ker And even if I comment `MultitenantAdminMixin` class I can't still see objects associated to other organizations, basically I don't see any difference between existing of `MultitenantAdminMixin` and lack of  `MultitenantAdminMixin` class. 
https://github.com/openwisp/openwisp-utils/blob/master/openwisp_utils/admin.py#L6

OltarzewskiK:
@nemesisdesign I know that, but I have created two books and two users which have permissions set to all and both are in other orgs and books are too in other orgs (first is in the same as user1 and second in the same as user2) , but user of org 1 can't see book of org2 even after commenting whole class.

atb00ker:
I am not sure if i understand the problem correctly, but i think one of your users might not be marked as `is_admin` can you confirm that? :smile:

OltarzewskiK:
@atb00ker They are both marked as `is_admin`:
screenshot from 2018-11-22 15-18-56

nemesisdesign:
same here, feel free to reopen if you intend to continue

PR #23, opened 2018-11-21
Added commit message style check

ppabcd:
Google Code-in Task : https://codein.withgoogle.com/dashboard/task-instances/5188396768034816/
- [x] prefix not present, eg: Done something (should be [qa] Done something)
- [x] no capital letter after prefix: [qa] done something
- [x] unnecessary final dot in first line, eg: [qa] Done something.
- [x] No blank line before first line and second line
- [x] Mentions an issue without closing it
- [x] Closes an issue in the long description but doesn't mention it in the short description

Fixes #20

Issue #20, opened 2018-11-18
[qa] Add commit message style check


nemesisdesign:
It's time to automate the commit message style checks according to [our guidelines](http://openwisp.io/docs/developer/contributing.html#commit-message-style-guidelines), that will make contributing a lot smoother for new contributors and less time consuming for reviewers.

These are the most common issues which we should address:

- prefix not present, eg: `Done something` (should be ``[qa] Done something``)
- no capital letter after prefix: `[qa] done something`
- unnecessary final dot in first line, eg: `[qa] Done something.`
- No blank line before first line and second line, eg:
```
[qa] Done something
Long description
```
- Mentions an issue without closing it, eg: `Implements #20` (won't close), `Closes #20 and #21` (won't close `#21` - but we should allow writing something like `Related to #33`, because sometimes we may want to reference an issue even though we are not solving it in that commit
- Closes an issue in the long description but doesn't mention it in the short description, eg:
```
[qa] Done something

Fixes #2
```
should be:
```
[qa] Done something #2

Fixes #2
```
(easier to read in the short log format)

This is not urgent but not unimportant because having a clean commit history is important for the long term sustainability of the project.
It doesn't make sense to keep enforcing these rules by hand.. it's clear to me that as the community grows we won't be able to do that, so automating this check is key.

We need something like the migration check implemented in #16.

ppabcd:
it's not clear.

PR #26, opened 2018-12-01
Improved our quality assurance checks

nemesisdesign:
@ppabcd see the improvements to your code, especially regarding readability

@atb00ker I found out a few things that needed improvement in the migration check as well, these kind of checks are designed to be used as linux commands so they should not raise exceptions in case of failure but simply print a message and exit with an exit code higher than zero.

I've added some basic docs in the readme and last but not least I've added a way to skip the commit message check (using the `#noqa` keyword in the commit message) in case we encounter some legit case which we didn't plan here and we don't want to block the build, we may remove this option in the future once we are confident all cases are covered.

Related to #20 and #23.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,6 @@ before_install:
   - pip install -U pip wheel setuptools
   - pip install --no-cache-dir -U -r requirements-test.txt
 
-# command to install requirements
 install:
   - pip install https://github.com/openwisp/openwisp-users/tarball/master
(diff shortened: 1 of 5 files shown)


nemesisdesign:
Damn, the travis build has gotten stuck and I cannot restart it. I will try to close this and re-open it.

nemesisdesign:
Didn't work. I will try to open a new one :angry:

PR #41, opened 2019-04-21
[admin] Removed logic duplicates #40

### LINKED ISSUE #40: Generalize easy to copy uuid, secret key and secret URL logic
**Reserved for GSoC Students**

We have different parts of different OpenWISP modules which have a `key` or `secret` field which has one or more of the following features:

- the key is automatically and randomly generated
  - [django-netjsonconfig](https://github.com/openwisp/django-netjsonconfig/blob/master/django_netjsonconfig/base/device.py#L21-L26)
  - [django-netjsongraph](https://github.com/netjson/django-netjsongraph/blob/master/django_netjsongraph/base/topology.py#L47-L53)
  - [openwisp-radius](https://github.com/openwisp/openwisp-radius/blob/master/openwisp_radius/models.py#L125-L127)
  - [openwisp-controller](https://github.com/openwisp/openwisp-controller/blob/master/openwisp_controller/config/models.py#L241-L247)
- the uuid field in the admin is readonly and can be copied easily:
  - [django-netjsonconfig](https://github.com/openwisp/django-netjsonconfig/blob/master/django_netjsonconfig/base/admin.py#L44) (look for `uuid` and look at the javascript file)
  - [openwisp-users](https://github.com/openwisp/openwisp-users/blob/master/openwisp_users/admin.py) (look for `uuid` and look at the javascript file)
- the key is used to build a URL which can be copied easily:
  - [django-netjsongraph](https://github.com/netjson/django-netjsongraph/blob/master/django_netjsongraph/base/admin.py#L41-L53) (look for `receive_url` and then find the `receive-url.js` javascript)

We need to:

- remove the duplication of this logic across modules and make it reusable
- add automated tests for this logic
- document its usage briefly in the README of this module
- update the other modules to use the generalized solution here in openwisp-utils

NoumbissiValere:
Removed the logics which were duplicated
in the others modules and redefined it here to enable reusability and ease debugging.

Fixes #40


NoumbissiValere:
@nemesisdesign Please can you can review this when you have time so that i make corrections if needed. I have equally done the modifications in the other `repos` as indicated at the issue's page. but travis test will fail for them if this one is not merge. 
Thanks.

tests/test_project/admin.py
```diff
@@ -24,9 +25,32 @@ class OperatorInline(admin.StackedInline):
...
+class ProjectAdmin(admin.ModelAdmin, UUIDAdmin):
     inlines = [OperatorInline]
     list_display = ['name']
+    readonly_fields = ['uuid']
+    fields = ['name', 'key', 'uuid']
+
+    class Media:
+        js = ('openwisp-utils/js/uuid.js',)
+
+    def _get_fields(self, fields, request, obj=None):
```
nemesisdesign: why are you adding this code and the lines below?

tests/test_project/admin.py
```diff
@@ -24,9 +25,32 @@ class OperatorInline(admin.StackedInline):
...
 
-class ProjectAdmin(admin.ModelAdmin):
+class ProjectAdmin(admin.ModelAdmin, UUIDAdmin):
     inlines = [OperatorInline]
     list_display = ['name']
+    readonly_fields = ['uuid']
+    fields = ['name', 'key', 'uuid']
+
+    class Media:
+        js = ('openwisp-utils/js/uuid.js',)
```
nemesisdesign: does it work also without adding `openwisp_utils` to `INSTALLED_APPS`? Can it actually be loaded in the browser correctly? Did you try running the test project in the browser and check this?

tests/test_project/admin.py
NoumbissiValere: I adding those lines so that the readonly fields (`uuid` in this case) should be should come up only in the change_view. without this, the `uuid` field will appear in the `add_view` with a None value. Equally this was the same approach used in `OrganizationAdmin` found in the `openwisp_users` repo.

NoumbissiValere:
@nemesisdesign Please can you look at the receive_url i added. i wish to know if it's ok before i add the tests.

openwisp_utils/static/openwisp-utils/js/receive_url.js
```diff
@@ -0,0 +1,10 @@
+django.jQuery(function ($) {
+    "use strict";
+     var p = $('.field-receive_url p, .field-receive_url > div > div'),
+            value = window.location.origin + p.text();
+     p.html('<input readonly id="id_receive_url" type="text" class="vTextField readonly" value="' + value + '">');
+     var field = $('#id_receive_url');
+     field.click(function () {
+        $(this).select();
+     });
+};
```
atb00ker: This needs to be `});` uneven brackets. :smile:

tests/test_project/admin.py
```diff
@@ -24,9 +25,32 @@ class OperatorInline(admin.StackedInline):
...
+        """
+        removes readonly_fields in add view
+        """
+        if obj:
+            return fields
+        new_fields = fields[:]
+        for field in self.readonly_fields:
+            if field in new_fields:
+                new_fields.remove(field)
+        return new_fields
```
atb00ker: Can't this be generalized and added to the `UUIDAdmin` class?

nemesisdesign: Correct, I think this should be generalized in the `UUIDAdmin` class.

nemesisdesign:
Thanks a lot @NoumbissiValere for your work, we merged your initial work after some improvements in PR #71. :+1:

Issue #39, opened 2019-02-13
[enchancement] Common CI script


atb00ker:
We need a common CI script to run all the CI tests, so that if there is minor change or a new check is added/removed, we do not need to update all the repositories with the minor change.

Read more: https://github.com/openwisp/openwisp-utils/pull/37#issuecomment-461733269

dwang:
Are these all the scripts that should be run?
- `checkcommit`
- `runcheckendline`
- `runcheckmigration`
- `runflake8`
- `runisort`

However, some of these checks don't apply to all of the repositories, how would you solve that?
For example, `runcheckmigration` wouldn't really apply in `openwisp-website`, but `checkcommit` and `runcheckendline` should still be run.

atb00ker:
@dwang I am thinking command line variable, by default, all tasks will run,
However, if i want to avoid `runcheckmigration`, i can simply do `--no-runcheckmigration` to avoid running that script, what do you think about this? :smile:

dwang:
Sounds good to me!

ankit-kumar-dwivedi:
@atb00ker this sounds great! are you working on this one?

atb00ker:
No. I am currently not working on it. :smile:

ankit-kumar-dwivedi:
okay I am working on it now

ankit-kumar-dwivedi:
@atb00ker in which order should the scripts run? 
I see `checkcommit` and `runcheckendline` should run on every repo.
So I am going to take a maximum of 3 arguments

atb00ker:
Tests are not dependant on each other so the order doesn't matter.
You can checkout the `.travis.yml` file in each repository to checkout what tests are running on that specific repository.
Checkout the same for the following repositories:
```
- django-ipam
- openwisp-ipam
- django-freeradius
- openwisp-freeradius
- openwisp-users
- openwisp-utils
- openwisp-config
- openwisp-controller
- django-loci
- django-netjsonconfig
- django-x509
```

ankit-kumar-dwivedi:
@atb00ker I see that the .travis.yml file in each repository is divided in 3 parts
- **before_install:** here `./runflake8` , `./runisort` , `./runcheckmigration` etc are called
- **install:** here are all install commands...
- **script:** `checkcommit` runs here and reposoitory specific test i.e jshint, ./runtest , `coverage` etc are called.
### Also there are different commands in each .travis.yml to install openwisp-utils
[...]

PR #45, opened 2019-12-05
Common CI script

waleko:
This is a pull request resolving issue #39.

New bash script `runchecks` replaces following scripts:

- `checkcommit`
- `runcheckendline`
- `runcheckmigration`
- `runflake8`
- `runisort`

It accepts `--skip-{check-name}` options (e.g. `./runchecks --skip-isort --skip-checkcommit`).
You can also run it with `--help` option (if you want to for some reason)

P.S. It's a GCI Task


runchecks
```diff
@@ -0,0 +1,115 @@
...
+            flag=1
+        fi
+    done
+    if [ $flag -ne 0 ]; then
+        exit $flag
+    fi
+}
+
+runcheckmigration() {
+    set -e
```
dwang: You can move `set -e` to the top and it will apply for the entire script.

runchecks
```diff
@@ -0,0 +1,115 @@
...
+    checkmigrations --migration-path ./tests/test_project/migrations/ || exit 1
+}
+
+runflake8() {
+    set -e
```
dwang: You can move `set -e` to the top and it will apply for the entire script.

runchecks
```diff
@@ -0,0 +1,115 @@
...
+    flake8 --max-line-length=110 \
+        --exclude=migrations,./tests/*settings*.py \
+        --ignore=W605,W504 || exit 1
+}
+
+runisort() {
+    set -e
```
dwang: You can move `set -e` to the top and it will apply for the entire script.

runchecks
```diff
@@ -0,0 +1,115 @@
...
+    isort --check-only --recursive --diff || exit 1
+}
+
+runcheckcommit() {
+    if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
+      # gets commit message of last commit before pull request merge
+      COMMIT_MESSAGE=$(git log $TRAVIS_PULL_REQUEST_SHA --format=%B -n 1)
+      printf "Checking commit message:\n\n"
+      printf "$COMMIT_MESSAGE\n\n"
+      checkcommit --message "$COMMIT_MESSAGE"
```
dwang: The arguments aren't being passed correctly. Can you take a look?
https://travis-ci.org/openwisp/openwisp-utils/jobs/621280059#L424

runchecks
```diff
@@ -0,0 +1,115 @@
...
+        fi
+    done
+    if [ $flag -ne 0 ]; then
+        exit $flag
+    fi
+}
+
+runcheckmigration() {
+    set -e
+    checkmigrations --migration-path ./tests/test_project/migrations/ || exit 1
```
dwang: The arguments aren't being passed correctly. Can you take a look?
https://travis-ci.org/openwisp/openwisp-utils/jobs/621280059#L427

waleko:
Hi, thanks for suggestions!

I've moved `set -e` to the main level. I also squashed my commits in one.

The thing with passing arguments turned up to be not the problem with my script. It's a problem with coverage. You can see some [old build](https://travis-ci.org/openwisp/openwisp-utils/jobs/525184823#L636) (even without `runchecks`) where these errors are still appearing. Presumably, `runchecks.py` runs `checkcommit` and `checkmigration` functions, that have no arguments to process which results in an error. I can take a deeper look into it if you wish.

If the check is failed the script will return non-zero exit code as you can see in [this build](https://travis-ci.org/openwisp/openwisp-utils/jobs/621277720#L426) when I myself ran into an issue of not following contributing guidelines on commit names (duh).

runchecks
waleko: The thing with passing arguments turned up to be not the problem with my script. It's a problem with coverage. You can see some [old build](https://travis-ci.org/openwisp/openwisp-utils/jobs/525184823#L636) (even without `runchecks`) where these errors are still appearing. Presumably, `runchecks.py` runs `checkcommit` and `checkmigration` functions, that have no arguments to process which results in an error. I can take a deeper look into it if you wish.

runchecks
waleko: The thing with passing arguments turned up to be not the problem with my script. It's a problem with coverage. You can see some [old build](https://travis-ci.org/openwisp/openwisp-utils/jobs/525184823#L636) (even without `runchecks`) where these errors are still appearing. Presumably, `runchecks.py` runs `checkcommit` and `checkmigration` functions, that have no arguments to process which results in an error. I can take a deeper look into it if you wish.

atb00ker:
Great work here! 

> @atb00ker thoughts about this?

I'd defer this to @nemesisdesign 

> Presumably, runchecks.py runs checkcommit and checkmigration functions, that have no arguments to process which results in an error. I can take a deeper look into it if you wish.

Yes, it's just the output while testing the scripts. :smile:


Is this ready for testing, shall I try it?

runchecks
```diff
@@ -0,0 +1,114 @@
...
+
+if ! $skip_checkmigration; then
+    runcheckmigration
+fi
+
+if ! $skip_flake8; then
+    runflake8
+fi
+
+if ! $skip_checkisort; then
```
atb00ker: This is not correct. Should be `$skip_isort` according to the variable above.

runchecks
```diff
@@ -0,0 +1,114 @@
+#!/bin/bash
+
+show_help() {
```
atb00ker: Just for the sake of prettification of the output, can you please try:
```
    printf "Usage:\n"
    printf "runchecks [--help] [--skip-checkendline] [--skip-checkmigration] [--skip-flake8] [--skip-isort] [--skip-checkcommit]\n"
    printf "\n\n"
    printf "Options:\n"
    printf "  --help\t\t\t: Show this help text\n"
    printf "  --skip-checkendline\t\t: Skip blank endline check\n"
    printf "  --skip-checkmigration\t\t: Skip migration name check\n"
    printf "  --skip-flake8\t\t\t: Skip flake8 check\n"
    printf "  --skip-isort\t\t\t: Skip isort check\n"
    printf "  --skip-checkcommit\t\t: Skip commit check\n"
```

waleko:
@atb00ker Made the changes

atb00ker:
Okay, I just realized a big mistake in the implementation.

The primary reason of of making this change is that in any repository we can simply install `openwisp-utils` and get our set quality assurance checks but I just realized that's not something we can do with shell script! (If you have a possible solution, please let me know, we can discuss it.)

Hence, I think you'll need to write this shell script in python instead as done for `checkcommit` here:
- https://github.com/openwisp/openwisp-utils/blob/master/openwisp_utils/qa.py#L74
- https://github.com/openwisp/openwisp-utils/blob/master/setup.py#L37

Now, we can install `openwisp-utils` and use `checkcommit` command to get our work done in repository.

waleko:
@atb00ker @dwang Thankfully, I found a much simpler workaround. I just added the script to `setup.py`.
```python
#... 
entry_points={
        'console_scripts': [
            'checkmigrations = openwisp_utils.qa:check_migration_name',
            'checkcommit = openwisp_utils.qa:check_commit_message',
        ],
    },
scripts=["runchecks"],
include_package_data=True,
#... 
```
Now when we call `pip install openwisp-utils` runchecks is installed automatically :smile:.
I think, it's a much simpler solution than rewriting it to python, as many of the checks themselves use command line, so with python it would be just a bit easier to write help page and to parse the arguments.

I also added `--migration-path <path>` argument to `runchecks`, as discussed with @atb00ker, and tweaked exit codes to be different for different checks and added `echo "Running {...} check"`, so it would be easier to understand which check failed.

dwang:
I'm okay with this, are you @atb00ker?
I'll do some testing right now and report back.

atb00ker:
@waleko That sounds great. It'll I'll try to test it later. For now, please try testing it:
- Install your openwisp_utils in another openwisp repository and try to make sure all the tests run with the command.

I have a couple of question:
1. If in future we decide to make this shell script as python script, would we need to change anything in repositories that it running on. (Example `django-loci` is going to use these these test, so, we will run `runchecks` in loci, but if we change it to python script, does anything change for `django-loci` repository to run the script)
2. Just confirming but it sounds like `--migration-path` is the only option you think is required from the old script but as a matter of fact we need other options like `quite`, `migrations_to_ignore` etc. Hope I didn't convey that incorrectly, `--migration-path` was only an example but we don't want to remove any of the features of the old scripts because of this new one! :smile:

atb00ker:
@dwang Sounds great, thanks! :smile:

waleko:
@dwang @atb00ker I hard-coded all parameters used by `checkcommit` and `checkmigrations`. I also "redisgned" help page, so it's easier to understand what parameters are for which check. 
```
Usage:
  runchecks --help
  runchecks {--migration-path <path> [--migrations-to-ignore <int>] | --skip-checkmigrations}
            {[--message <commit-message>] | --skip-checkcommit}
            [--skip-checkendline]
            [--skip-flake8]
            [--skip-isort]
            [--quiet]

Runs all checks for quality assurance. Unneeded checks may be skipped by passing --skip-<check-name> argument.

General options:
  --help                        : Show this help text
  --quiet                       : Suppress output
Migration check options:
  --migration-path <path>       : Path to migrations/ folder
  --migrations-to-ignore <int>  : Number of migrations after which checking of migration file names should begin,
                                  say, if checking needs to start after 0003_auto_20150410_3242.py value should be 3
  --skip-checkmigrations        : Skip migration name check
Commit name check options:
  --message <commit-message>    : Commit message (by default is already equal to $COMMIT_MESSAGE)
  --skip-checkcommit            : Skip commit check
Blank endlines check options:
  --skip-checkendline           : Skip blank endlines check
Flake8 check options:
  --skip-flake8                 : Skip flake8 check
Isort check options:
  --skip-isort                  : Skip isort check
```
Now if we change `runchecks` to a python script, all we would need to do is to parse all arguments and pass the arguments object to an auxiliary function and it will just use its arguments 😄. The only limitation to it is that all arguments names have to be unique, but I think, that it's still a good thing, because it would be better having a more descriptive argument name.

I also added a `--quiet` parameter. It suppresses `runcheck` output and output of all auxiliary programs as well.

nemesisdesign:
@dwang 

> I also noticed that all the checks are run at the beginning now, but maybe it might be better to run them all at the end since code changes will not be tested if something minor like the commit message check fails. @nemesisdesign @atb00ker thoughts about this?

If something is wrong, it is better to fail as early as possible so that who's working on the PR finds out early and can fix quickly.

runchecks
```diff
@@ -0,0 +1,188 @@
...
+  printf "Usage:\n"
```
nemesisdesign: when the final `\n` is needed, why can't you just use `echo` instead? I think that is cleaner. Unless you have a better reason to use `printf` I ask you to use `echo`.

.travis.yml
```diff
@@ -32,21 +32,10 @@ install:
   - pip install -e .[users] .[qa]
 
 before_script:
-  - ./runflake8
-  - ./runisort
-  - ./runcheckmigration
-  - ./runcheckendline
+  - runchecks --migration-path "./tests/test_project/migrations"
```
nemesisdesign: let's rename this script: `openwisp-utils-qa-checks`, this way we will make it obvious where the script comes from and contributors will quickly be able to find it

runchecks
```diff
@@ -0,0 +1,188 @@
...
+      checkcommit --message "$COMMIT_MESSAGE"
+    else
+      checkcommit --message "$COMMIT_MESSAGE" --quiet
+    fi
+  fi
+}
+
+## Main
+
+skip_checkendline=false
```
nemesisdesign: please let's stay consistent and use uppercase name for these variables

runchecks
waleko: `echo` doesn't work so well with `\t`. Sometimes it just prints "\t" in plaintext, so I changed it to `printf`, so it has a more stable behavior.

setup.py
```diff
@@ -37,6 +37,7 @@
             'checkcommit = openwisp_utils.qa:check_commit_message',
         ],
     },
+    scripts=["openwisp-utils-qa-checks"],
```
nemesisdesign: please use single quotes (`'`) for consistency, keep in mind to always be consistent with the styles already used in the program

PR #48, opened 2019-12-10
[enhancement] Overrided django.contrib.admin.AdminSite #42

### LINKED ISSUE #42: Override django.contrib.admin.AdminSite instead of modifying it at runtime
The function openwisp_admin() in openwisp_utils.admin_theme.admin changes some settings of the vanilla django admin site at runtime. This makes it difficult to use an overridden admin site instead of the default one together with OpenWISP.

It would be better to override the vanilla admin site instead: https://docs.djangoproject.com/en/2.2/ref/contrib/admin/#customizing-the-adminsite-class Then, for further customization, this overridden class can be overridden once more.

oltarzewskik:
Fixes #42

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -99,23 +99,17 @@ Using the utilities in OpenWISP modules
 Using the ``admin_theme``
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
-For using the customized admin theme,
-
-* Make sure you've added ``openwisp_utils.admin_theme`` to ``INSTALLED_APPS``
-  in ``settings.py``
-* Add the following into your ``urls.py`` file which contains ``admin`` urls.
+* Add ``openwisp_utils.admin_theme`` to ``INSTALLED_APPS`` in ``settings.py``.
+* Import ``admin_site`` and make changes like following into your ``urls.py`` file which contains ``admin`` urls:
 
 .. code-block:: python
 
-    from django.conf.urls import include, url
-
-    from openwisp_utils.admin_theme.admin import admin, openwisp_admin
-
-    openwisp_admin()
+    from django.conf.urls import url
+    from openwisp_utils.admin_theme.site import admin_site
 
     urlpatterns = [
         # other url patterns
-        url(r'^admin/', include(admin.site.urls)),
+        url(r'^admin/', admin_site.urls),
     ]
 
 Admin mixins
@@ -210,6 +204,13 @@ template ``context_processors`` in ``settings.py`` as shown below.
 
 Settings
 ^^^^^^^^
+
+``OPENWISP_ADMIN_SITE_CLASS``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**default**: ``openwisp_utils.admin_theme.admin.OpenwispAdminSite``
+
+If you need to use a customized admin site class, you can use this setting.
 
 ``OPENWISP_ADMIN_SITE_TITLE``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/openwisp_utils/admin_theme/admin.py
+++ b/openwisp_utils/admin_theme/admin.py
@@ -3,20 +3,29 @@ from django.contrib import admin
 from django.utils.translation import ugettext_lazy
 
 
-def openwisp_admin(site_url=None):
+class OpenwispAdminSite(admin.AdminSite):
     # <title>
-    admin.site.site_title = getattr(settings,
-                                    'OPENWISP_ADMIN_SITE_TITLE',
-                                    'OpenWISP Admin')
-    # link to frontend
-    admin.site.site_url = site_url
+    site_title = getattr(settings,
+                         'OPENWISP_ADMIN_SITE_TITLE',
+                         'OpenWISP Admin')
     # h1 text
-    admin.site.site_header = getattr(settings,
-                                     'OPENWISP_ADMIN_SITE_HEADER',
-                                     'OpenWISP')
+    site_header = getattr(settings,
+                          'OPENWISP_ADMIN_SITE_HEADER',
+                          'OpenWISP')
     # text at the top of the admin index page
-    admin.site.index_title = ugettext_lazy(
+    index_title = ugettext_lazy(
         getattr(settings,
                 'OPENWISP_ADMIN_INDEX_TITLE',
                 'Network administration')
     )
+
+
+def openwisp_admin(site_url=None):
+    """
+    openwisp_admin function is deprecated and discouraged to use,
+    you should use admin_site object from openwisp_utils.admin_theme.site instead
+    """
+    admin.site.site_title = OpenwispAdminSite.site_title
+    admin.site.site_url = site_url  # link to frontend
+    admin.site.site_header = OpenwispAdminSite.site_header
+    admin.site.index_title = OpenwispAdminSite.index_title
--- a/tests/settings.py
+++ b/tests/settings.py
@@ -77,6 +77,8 @@ DATABASES = {
     }
 }
 
+OPENWISP_ADMIN_SITE_CLASS = 'test_project.site.CustomAdminSite'
+
 SITE_ID = '1'
 EMAIL_PORT = '1025'
 LOGIN_REDIRECT_URL = 'admin:index'
--- a/tests/test_project/admin.py
+++ b/tests/test_project/admin.py
@@ -1,14 +1,17 @@
 from django.contrib import admin
 from django.forms import ModelForm
 from openwisp_utils.admin import AlwaysHasChangedMixin, ReadOnlyAdmin
+from openwisp_utils.admin_theme.site import admin_site
 
 from .models import Operator, Project, RadiusAccounting
 
 
+@admin.register(Operator, site=admin_site)
 class OperatorAdmin(admin.ModelAdmin):
     list_display = ['first_name', 'last_name']
 
 
+@admin.register(RadiusAccounting, site=admin_site)
 class RadiusAccountingAdmin(ReadOnlyAdmin):
     list_display = ['session_id', 'username']
     fields = ['session_id', 'username']
@@ -24,11 +27,7 @@ class OperatorInline(admin.StackedInline):
     extra = 0
 
 
+@admin.register(Project, site=admin_site)
 class ProjectAdmin(admin.ModelAdmin):
     inlines = [OperatorInline]
     list_display = ['name']
-
-
-admin.site.register(Operator, OperatorAdmin)
-admin.site.register(RadiusAccounting, RadiusAccountingAdmin)
-admin.site.register(Project, ProjectAdmin)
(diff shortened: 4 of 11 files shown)


oltarzewskik:
@okraits @atb00ker @nemesisdesign After substituting `admin.site` in `urls.py` by overrided `AdminSite` object (`admin_site`) every `/admin/openwisp_users` URLs are missing.

Could you give me some advice why does it happens?

openwisp_utils/admin_theme/admin.py
```diff
@@ -20,3 +20,9 @@ def openwisp_admin(site_url=None):
                 'OPENWISP_ADMIN_INDEX_TITLE',
                 'Network administration')
     )
+
+
+class ModifiedAdminSite(admin.AdminSite):
+    index_title = ugettext_lazy(getattr(settings, 'OPENWISP_ADMIN_INDEX_TITLE', 'Network administration'))
```
nemesisdesign: follow indentation according to how it was done previously

tests/test_project/admin.py
```diff
@@ -29,6 +30,6 @@ class ProjectAdmin(admin.ModelAdmin):
     list_display = ['name']
 
 
-admin.site.register(Operator, OperatorAdmin)
-admin.site.register(RadiusAccounting, RadiusAccountingAdmin)
-admin.site.register(Project, ProjectAdmin)
+admin_site.register(Operator, OperatorAdmin)
```
nemesisdesign: should be done this way: https://docs.djangoproject.com/en/3.0/ref/contrib/admin/#django.contrib.admin.register

openwisp_utils/admin_theme/apps.py
```diff
@@ -0,0 +1,4 @@
+from django.contrib.admin.apps import AdminConfig
+
+class ModifiedAdminConfig(AdminConfig):
```
nemesisdesign: rename to `OpenWispAdminConfig`

tests/settings.py
```diff
@@ -71,7 +71,7 @@
                 'django.template.context_processors.request',
                 'django.contrib.auth.context_processors.auth',
                 'django.contrib.messages.context_processors.messages',
-                'openwisp_utils.admin_theme.context_processor.menu_items'
+                # 'openwisp_utils.admin_theme.context_processor.menu_items'
```
oltarzewskik: I commented this line because without changes to `openwisp-users` that line will lead to errors

oltarzewskik:
@okraits @nemesisdesign @atb00ker what about docs?

openwisp_utils/admin_theme/admin.py
```diff
@@ -20,3 +22,22 @@ def openwisp_admin(site_url=None):
                 'OPENWISP_ADMIN_INDEX_TITLE',
                 'Network administration')
     )
+
```
nemesisdesign: change also `openwisp_admin` so that we avoid the duplication of lines like:

`site_title = getattr(settings, # ...`.

Move the function below, and use the class attributes defined in `OpenwispAdminSite` instead of redefining them.

Add a docstring to the function explaining that its usage it's deprecated.

openwisp_utils/settings.py
```diff
@@ -1,3 +1,4 @@
 from django.conf import settings
 
 EXTENDED_APPS = getattr(settings, 'EXTENDED_APPS', [])
+CUSTOM_ADMIN_SITE = getattr(settings, 'CUSTOM_ADMIN_SITE')
```
nemesisdesign: move this to a new file in `admin_theme/settings.py`, since this is a setting of the admin_theme sub app and not the general openwisp-utils.

let's call this setting `OPENWISP_ADMIN_SITE_CLASS`.
Yes you will have to document this new feature.

You have to provide the default value of this setting, the default value has to be the path to the `OpenwispAdminSite` class

tests/settings.py
```diff
@@ -26,7 +26,7 @@
     # test project
     'test_project',
     # admin
-    'django.contrib.admin',
+    'openwisp_utils.admin_theme.apps.OpenwispAdminConfig',
```
nemesisdesign: this change should not be needed, also because `openwisp_utils.admin_theme` is already called earlier:
https://github.com/openwisp/openwisp-utils/blob/9f065e3d3be5c84f115cd3b4f9744c5e55fd7feb/tests/settings.py#L17

tests/settings.py
nemesisdesign: we need this, let the errors show up so we can understand what's going on

openwisp_utils/admin_theme/apps.py
```diff
@@ -0,0 +1,5 @@
+from django.contrib.admin.apps import AdminConfig
+from ..settings import CUSTOM_ADMIN_SITE
+
+class OpenwispAdminConfig(AdminConfig):
+    default_site = CUSTOM_ADMIN_SITE
```
nemesisdesign: don't do this, revert the changes in this file and follow my subsequent suggestions

openwisp_utils/admin_theme/admin.py
```diff
@@ -20,3 +22,22 @@ def openwisp_admin(site_url=None):
...
+                          'OPENWISP_ADMIN_SITE_HEADER',
+                          'OpenWISP')
+    # text at the top of the admin index page
+    index_title = ugettext_lazy(
+        getattr(settings,
+                'OPENWISP_ADMIN_INDEX_TITLE',
+                'Network administration')
+    )
+
+openwisp_admin_site = import_string(apps.get_app_config('admin').default_site)()
```
nemesisdesign: Move this to a separate file called `site.py` and do something like:

```python
from . import settings as app_settings

admin_site_class = import_string(app_settings.OPENWISP_ADMIN_SITE_CLASS)
admin_site = admin_site_class()
```

This way we avoid possible circular import issues.

tests/settings.py
```diff
@@ -84,6 +84,7 @@
     }
 }
 
+CUSTOM_ADMIN_SITE = 'openwisp_utils.admin_theme.admin.OpenwispAdminSite'
```
nemesisdesign: please define a custom site class in the `test_project`, which extends `OpenwispAdminSite` and adds a custom attribute, then write a test which asserts the presence of this custom attribute.

Use this settings file to specify the path to the custom admin site class.

tests/test_project/admin.py
```diff
@@ -24,11 +29,8 @@ class OperatorInline(admin.StackedInline):
     extra = 0
 
 
+@admin.register(Project, site=admin_site)
+@admin.register(Project, site=admin.site)
```
oltarzewskik: it do pretty much the same as `admin.site.register(Project, ProjectAdmin)`

tests/settings.py
oltarzewskik: I will uncomment when I will have tests working

README.rst
```diff
@@ -106,23 +106,38 @@ Using the utilities in OpenWISP modules
...
+
+    from django.conf.urls import url
+    from openwisp_utils.admin_theme.site import admin_site
+
+    urlpatterns = [
+        # other url patterns
+        url(r'^admin/', admin_site.urls),
+    ]
+
+Alternatively you can use ``openwisp_admin()`` function, but its usage is discouraged.
```
nemesisdesign: do not mentione this anymore

README.rst
```diff
@@ -218,6 +233,13 @@ template ``context_processors`` in ``settings.py`` as shown below.
 Settings
 ^^^^^^^^
 
+``OPENWISP_ADMIN_SITE_CLASS``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**default**: ``openwisp_utils.admin_theme.admin.OpenwispAdminSite``
+
+Path to custom admin site class
```
nemesisdesign: Change to:

`If you need to use a customized admin site class, you can use this setting.`

README.rst
```diff
@@ -106,23 +106,38 @@ Using the utilities in OpenWISP modules
...
 * Add the following into your ``urls.py`` file which contains ``admin`` urls.
 
 .. code-block:: python
 
-    from django.conf.urls import include, url
+    from django.conf.urls import url
 
     from openwisp_utils.admin_theme.admin import admin, openwisp_admin
 
     openwisp_admin()
```
nemesisdesign: remove this

oltarzewskik: @nemesisdesign I have to remove whole section about `openwisp_admin` from README, right?

nemesisdesign: what section?

atb00ker:
I might have misunderstood something here, but the fact that we need to change `openwisp_users` to fix testcases of `openwisp-utils` sounds like a coupling problem. 
( So far, I haven't understood why it is this way, I'll investigate again later! :smile: )

oltarzewskik:
@atb00ker I think that I know how to fix users issue , will make changes :+1:

README.rst
oltarzewskik: @nemesisdesign I already made changes in `README` is it now like supposed to be?

nemesisdesign:
> I might have misunderstood something here, but the fact that we need to change `openwisp_users` to fix testcases of `openwisp-utils` sounds like a coupling problem.
> ( So far, I haven't understood why it is this way, I'll investigate again later!  )

You're right! I did not think about this.

It's used during tests, but I'm not sure this is really required anymore. Before there was some overlap, some multitenancy features were shipped here and tested using openwisp-users. I remember that in the last year we have moved everything to openwisp-users so I'm not 100% we really need to have openwisp-users anymore here. If we don't need it I'm totally in favour of removing it! :+1:

oltarzewskik:
@nemesisdesign If we will be sure that `openwisp-users` is not longer needed, I am willing to remove it from `openwisp-utils` :+1:

tests/test_project/tests/test_admin.py
```diff
@@ -53,3 +53,14 @@ def test_alwayshaschangedmixin(self):
         self.assertEqual(Operator.objects.count(), 1)
         project = Project.objects.first()
         self.assertEqual(project.name, params['name'])
+
+    def test_custom_admin_site(self):
+        url = reverse('admin:logout')
+        response = self.client.get(url)
+        content = str(response.content)
+        # Login again for next tests
+        user = User.objects.get(username='administrator')
```
atb00ker: In my sample I used `logout` template just because I don't have to do a lot of work.
But for the final work, can you please use something like `password_change_template` to avoid this relogin? 

Thanks.

openwisp_utils/admin_theme/admin.py
```diff
@@ -3,20 +3,27 @@
...
+    index_title = ugettext_lazy(
         getattr(settings,
                 'OPENWISP_ADMIN_INDEX_TITLE',
                 'Network administration')
     )
+
+
+# openwisp_admin function is deprecated and discouraged to use,
+# you should use admin_site object from openwisp_utils.admin_theme.site instead
+def openwisp_admin(site_url=None):
```
atb00ker: Small nitpick, since federico commented: "Add a docstring to the function explaining that its usage it's deprecated."

Can you please change this to:
```
def openwisp_admin(site_url=None):
"""
openwisp_admin function is deprecated and discouraged to use,
you should use admin_site object from openwisp_utils.admin_theme.site instead
"""
``` 

docstrings and comments have a small differences, some documenation parsers read the docstring to explain the user important notice / usage about the function.
I am not sure if that'll be important in this case but can you still make this change?

PR #51, opened 2019-12-11
Common CI script: make all checks even in case of failures

waleko:
This pull request makes `openwisp-utils-qa-checks` run all checks, even if some fail. The exit code's _i_-th bit represents whether corresponding check has failed or not. Here's a table of current checks' errors:

| No. | Name | Raised at |
|-----|--------------------------------------------------------------------------------|-----------------------------|
| 0 | `checkendline` fail | openwisp-utils-qa-checks:50 |
| 1 | `checkmigrations` fail | openwisp-utils-qa-checks:61 |
| 2 | `flake8` fail | openwisp-utils-qa-checks:66 |
| 3 | `isort` fail | openwisp-utils-qa-checks:71 |
| 4 | No `migration-path` provided and `--skip-checkmigrations` option is not chosen | openwisp-utils-qa-checks:58 |
| 5 | `checkcommit` fail | openwisp-utils-qa-checks:80 |

### DIFF
--- a/openwisp-utils-qa-checks
+++ b/openwisp-utils-qa-checks
@@ -29,6 +29,8 @@ show_help() {
   printf "  --skip-isort\t\t\t: Skip isort check\n"
 }
 
+echoerr() { echo "$@" 1>&2; }
+
 runcheckendline() {
   echo "Running blank endline check"
   flag=0
@@ -42,36 +44,46 @@ runcheckendline() {
     fi
   done
   if [ $flag -ne 0 ]; then
-    exit 1
+    echoerr "ERROR: Blank endline check failed!"
+    FAILURE=1
+  else
+    echo "SUCCESS: Blank endline check successful!"
   fi
 }
 
 runcheckmigrations() {
   echo "Running migration name check"
   if [ "$MIGRATION_PATH" == "" ]; then
-    echo "No migration path specified!"
-    exit 5
+    echoerr "ERROR: No migration path specified!"
+    FAILURE=1
+    return
   fi
-  checkmigrations --migration-path "$MIGRATION_PATH" --migrations-to-ignore "$MIGRATIONS_TO_IGNORE" || exit 2
+  checkmigrations --migration-path "$MIGRATION_PATH" --migrations-to-ignore "$MIGRATIONS_TO_IGNORE" \
+    && echo "SUCCESS: Migration name check successful!" \
+    || { echoerr "ERROR: Migration name check failed!"; FAILURE=1; }
 }
 
 runflake8() {
-  echo "Running flake8 check"
-  flake8 || exit 3
+  flake8 && echo "SUCCESS: Flake8 check successful!" \
+    || { echoerr "ERROR: Flake8 check failed!"; FAILURE=1; }
 }
 
 runisort() {
-  echo "Running isort"
-  isort --check-only --recursive --diff || exit 4
+  echo "Running isort check"
+  isort --check-only --recursive --diff \
+    && echo "SUCCESS: Isort check successful!" \
+    || { echoerr "ERROR: Isort check failed!"; FAILURE=1; }
 }
 
 runcheckcommit() {
-  echo "Running checkcommit"
   [ -n "$TRAVIS_PULL_REQUEST" ] || TRAVIS_PULL_REQUEST="false"
   if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
-    printf "Checking commit message:\n\n"
-    printf "%s\n\n" "$COMMIT_MESSAGE"
-    checkcommit --message "$COMMIT_MESSAGE" || exit 6
+    echo "Running commit message check"
+    checkcommit --message "$COMMIT_MESSAGE" \
+    && echo "SUCCESS: Commit name check successful!" \
+    || { printf "Checking commit message:\n\n"; printf "%s\n\n" "$COMMIT_MESSAGE"; echoerr "ERROR: Commit name check failed!"; FAILURE=1; }
+  else
+    echo "SUCCESS: Commit name check successful!"
   fi
 }
 
@@ -85,7 +97,7 @@ SKIP_CHECKCOMMIT=false
 MIGRATION_PATH=""
 MIGRATIONS_TO_IGNORE="0"
 COMMIT_MESSAGE=$(git log $TRAVIS_PULL_REQUEST_SHA --format=%B -n 1)
-
+FAILURE=0
 
 while [ "$1" != "" ]; do
   case "$1" in
@@ -151,3 +163,5 @@ fi
 if ! $SKIP_CHECKCOMMIT; then
   runcheckcommit
 fi
+
+exit $FAILURE


waleko:
@nemesisdesign Thanks for the suggestions. I don't really understand why different exit codes are a bad idea, maybe because of the code complexity, but I don't actually think it's that complicated. 

Anyways, I removed all exit code differences. Now if check is failed it will print out a message saying, that the check was failed and receive additional details  from an auxiliary program (e.g. `checkcommit`, `flake8`, etc) decribing the exact problem (well, the same as previously).

waleko:
Thanks for the suggestions. The travis build failing isn't because of this commit, I noticed already at the start, it is caused by commit [bc18f4a](https://github.com/openwisp/openwisp-utils/commit/bc18f4a01e77c759d6d96193ee00a264db098834), you can see it in this [master branch build](https://travis-ci.org/openwisp/openwisp-utils/jobs/623191429#L386).

Thank you for pointing out, that info message appear in some random way. I messed up && || statements, now it's fixed.

I also made some more simplifications, so the code would be simpler and easier to read. I also changed those info messages to something of your style (`SUCCESS:` and `ERROR:`)

openwisp-utils-qa-checks
```diff
@@ -151,3 +162,5 @@ fi
 if ! $SKIP_CHECKCOMMIT; then
   runcheckcommit
 fi
+
+exit $FAILURE
```
atb00ker: Can you please change it to 
```
    if [[ $TRAVIS = 1 ]]; then
        exit $FAILURE
    fi
```

Why? -- Because if I am running these tests on my personal machine, most likely I don't want to exit after the tests. (It'll be quite irritating to open another terminal and activate `venv` again and again, wouldn't you agree? :joy: )

openwisp-utils-qa-checks
```diff
@@ -42,36 +43,45 @@ runcheckendline() {
...
+    && echo "SUCCESS: Isort check successful!" \
+    || { echoerr "FAIL: Isort check failed!"; FAILURE=1; }
 }
 
 runcheckcommit() {
-  echo "Running checkcommit"
   [ -n "$TRAVIS_PULL_REQUEST" ] || TRAVIS_PULL_REQUEST="false"
   if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
     printf "Checking commit message:\n\n"
     printf "%s\n\n" "$COMMIT_MESSAGE"
```
atb00ker: Can you please move it inside the `FAIL` condition only? 

Why? (It's a nitpick really!)
- Because all the other tests are giving additional output only when there is a failure, not with success, so I think we should keep it uniform.
- Unix philosophy of programming which says that the program shouldn't give an output to user unless the output is required and I think that in success case, I don't need to see the commit message, that's all!

openwisp-utils-qa-checks
atb00ker: Actually, I'll resolve this one because the script can run without exiting the terminal. 
( Sorry, I was thinking about the `docker-openwisp` script which required this when I was making it because I needed to import those test functions to run them during development! )

waleko:
@atb00ker I squashed the commits. 

Don't really know why `flake8` works flawlessly now. I guess, it might have something to do with [`ea84225`](https://github.com/openwisp/openwisp-utils/commit/ea84225a014fa088686ce889a86bb0c87635dfb2) commit, though I'm not sure.

Also, for some bizarre reason _coverage_ stalls for this PR every time. Not really vital here, but not very great.

atb00ker:
> Don't really know why `flake8` works flawlessly now. I guess, it might have something to do with ea84225 commit, though I'm not sure.

I think it's not rebased to `master`. Can you please do that and check?

waleko:
@atb00ker It seams, that the commit, that I'm talking about ([`ea84225`](https://github.com/openwisp/openwisp-utils/commit/ea84225a014fa088686ce889a86bb0c87635dfb2)) is on master branch.

I now know the reason why this happens. It is indeed because of that commit. I got to this by comparing my own travis build for my forked repo which [failed](https://travis-ci.com/waleko/openwisp-utils/jobs/267666244#L432) and this repo's build which [succeeded](https://travis-ci.org/openwisp/openwisp-utils/jobs/625719906). They both ran the exact same commit. The only difference between them is that master branch of `openwisp/openwisp-utils` has a fix for `flake8` in `setup.cfg`. I then looked into, how travis clones the repo. It executes following steps:
```bash
$ git clone --depth=50 https://github.com/openwisp/openwisp-utils.git openwisp/openwisp-utils
$ cd openwisp/openwisp-utils
$ git fetch origin +refs/pull/51/merge:
$ git checkout -qf FETCH_HEAD
```
As you can see it at first cloning master branch and then fetching PR's branch. I replicated these commands on my machine and I got the modified `setup.cfg` file! Then `flake8`, that is installed using pip presumably gets it config from that file. So it means, that **commits on master indeed affect PR's builds**.

openwisp-utils-qa-checks
```diff
@@ -29,8 +29,9 @@ show_help() {
   printf "  --skip-isort\t\t\t: Skip isort check\n"
 }
 
+echoerr() { echo "$@" 1>&2; }
+
 runcheckendline() {
-  echo "Running blank endline check"
```
nemesisdesign: I would not remove this line, I think it's useful to have it.

openwisp-utils-qa-checks
```diff
@@ -42,36 +43,43 @@ runcheckendline() {
...
-    exit 1
+    echoerr "ERROR: Blank endline check failed!"
+    FAILURE=1
+  else
+    echo "SUCCESS: Blank endline check successful!"
   fi
 }
 
 runcheckmigrations() {
-  echo "Running migration name check"
```
nemesisdesign: I would not remove this line, I think it's useful to have it.

openwisp-utils-qa-checks
```diff
@@ -86,6 +94,7 @@ MIGRATION_PATH=""
 MIGRATIONS_TO_IGNORE="0"
 COMMIT_MESSAGE=$(git log $TRAVIS_PULL_REQUEST_SHA --format=%B -n 1)
```
nemesisdesign: I would remove this blank line

openwisp-utils-qa-checks
```diff
@@ -42,36 +43,43 @@ runcheckendline() {
...
+    echo "SUCCESS: Blank endline check successful!"
   fi
 }
 
 runcheckmigrations() {
-  echo "Running migration name check"
   if [ "$MIGRATION_PATH" == "" ]; then
-    echo "No migration path specified!"
-    exit 5
+    echoerr "No migration path specified!"
```
nemesisdesign: looks like `ERROR: ` is missing here

openwisp-utils-qa-checks
```diff
@@ -42,36 +43,43 @@ runcheckendline() {
...
+    FAILURE=1
+    return
   fi
-  checkmigrations --migration-path "$MIGRATION_PATH" --migrations-to-ignore "$MIGRATIONS_TO_IGNORE" || exit 2
+  checkmigrations --migration-path "$MIGRATION_PATH" --migrations-to-ignore "$MIGRATIONS_TO_IGNORE" \
+    && echo "SUCCESS: Migration name check successful!" \
+    || { echoerr "FAIL: Migration name check failed!"; FAILURE=1; }
```
nemesisdesign: why did you use `FAIL:` here while in the others error messages you used `ERROR:`? 
Let's keep consistency, I like `ERROR: ` because it's very explicit.

openwisp-utils-qa-checks
```diff
@@ -42,36 +43,43 @@ runcheckendline() {
...
 
 runflake8() {
-  echo "Running flake8 check"
-  flake8 || exit 3
+  flake8 && echo "SUCCESS: Flake8 check successful!" \
+    || { echoerr "FAIL: Flake8 check failed!"; FAILURE=1; }
 }
 
 runisort() {
-  echo "Running isort"
```
nemesisdesign: Leave the line, but write `Running isort check` please

openwisp-utils-qa-checks
```diff
@@ -42,36 +43,43 @@ runcheckendline() {
...
+  isort --check-only --recursive --diff \
+    && echo "SUCCESS: Isort check successful!" \
+    || { echoerr "FAIL: Isort check failed!"; FAILURE=1; }
 }
 
 runcheckcommit() {
-  echo "Running checkcommit"
   [ -n "$TRAVIS_PULL_REQUEST" ] || TRAVIS_PULL_REQUEST="false"
   if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
-    printf "Checking commit message:\n\n"
```
nemesisdesign: Leave this line, but use `Running commit message check`. please.

waleko:
@nemesisdesign Made minor fixes. Couldn't figure out how to rebase onto master, as I don't have write permissions. I also don't understand, why do we need to rebase at all if we're going to merge this commit anyways.

dwang:
If you check your fork, it says `This branch is 1 commit ahead, 2 commits behind openwisp:master`. We want it to just say 1 commit ahead.  Rebasing on master means pulling the latest commits from openwisp:master into your fork and reapplying the commits that you made on top of the commits from openwisp:master.

Also, I noticed that you are making this PR from the master branch on your fork. Instead, you should be making a separate branch on your fork for the PR. Take a look [here](http://openwisp.io/docs/developer/contributing.html#how-to-commit-your-changes-properly).

Everything else looks good to me. Good job! 👍

waleko:
@dwang @nemesisdesign @atb00ker I successfully rebased on master. I agree, that having a descriptive source branch name is important, but changing the branch now will result in closing this PR, which is perhaps not a good idea. So I guess, the best and easiest solution is to just let the branch be `waleko:master`...

nemesisdesign:
There's more small issues related to this topic: https://github.com/openwisp/openwisp-utils/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+openwisp-qa-check

It would be great to solve all of them so then we can go ahead and implement this check in all the modules.

Issue #47, opened 2019-12-10
[docs] Document openwisp-utils-qa-check


nemesisdesign:
Add the description of `checkendline` and `openwisp-utils-qa-check` to the README, in a similar way to how the other quality assurance scripts are documented.

TPath123:
Unlike the other individual QA checks, `checkendline` does not seem to be a command-line tool (https://github.com/openwisp/openwisp-utils/blob/master/setup.py#L35) nor is it in the [qa.py](https://github.com/openwisp/openwisp-utils/blob/master/openwisp_utils/qa.py) file

It only appears in the main `openwisp-utils-qa-check` file as a check, is this what you want me to document?

nemesisdesign:
@TPath123 you're right, the point of the issue is that we need to document the features and usage of `openwisp-utils-qa-check`, which has been introduced without being documented. 
A bit of documentation has just been added in #60, but we need to add a general description of what it does, its purpose, its features.

PR #60, opened 2019-12-21
[ci] Add support for multiple check migrations

### LINKED ISSUE #52: [ci] Improve checkmigrations in openwisp-utils-qa-checks
For now, checkmigrations in the new `openwisp-utils-qa-checks` common CI script is supported only for **one migration path and one migration-to-ignore**.
In some repo (like [django-ipam](https://github.com/openwisp/django-ipam), [django-loci](https://github.com/openwisp/django-loci), and more) it's required to check on **two migration-path**.
And even some repo (like [openwisp-users](https://github.com/openwisp/openwisp-users)) it's required to check on **two migration-path** and **two different migrations-to-ignore**.
To deal with this, we can:
1. Still using the custom `runcheckmigration` in those repo and use `openwisp-utils-qa-checks` with --skip-checkmigration. The disadvantage is, well, why should checkmigration exist in the common CI test then... :sweat_smile: 
2. The 'hacky' ways by using two lines of `openwisp-utils-qa-checks`, with the second call to check for other checkmigration and skip all tests. The disadvantage is when there is an addition/removal in the CI test, we should change those repositories too.

Either ways are fine, but kind of defeated the purpose of `openwisp-utils-qa-checks` that simplify the workflow of minor change in the CI tests without having to deal with all repositories. So I suggest that we might support multiple checkmigration, separated by semicolon. For example, we can do this for [openwisp-users](https://github.com/openwisp/openwisp-users/blob/master/runcheckmigration)

```
openwisp-utils-qa-checks --migration-path "./openwisp_users/migrations/;./tests/testapp/migrations/" --migrations-to-ignore "3;2"
```

Feel free to discuss/suggest/give your opinion about the better ways to do this :smile:

marfgold1:
Add support for multiple check migrations by passing the arguments with space delimited string.
Closes #52 

For example:
```
checkmigrations --migrations-to-ignore 3 \
                --migration-path ./openwisp_users/migrations/ || exit 1

checkmigrations --migrations-to-ignore 2 \
                --migration-path ./tests/testapp/migrations/ || exit 1
```
Can be changed by:
```
openwisp-utils-qa-checks --migration-path "./openwisp_users/migrations/ ./tests/testapp/migrations/" --migrations-to-ignore "3 2"
```

If `MIGRATIONS_TO_IGNORE` were not set, or the length of it's array is less than the length of `MIGRATION_PATH` array, then `MIGRATIONS_TO_IGNORE` will be set to 0 by default.
I also add progress info for each check to make it easier for us to debug later.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -267,6 +267,23 @@ Quality Assurance checks
 This package contains some common QA checks that are used the
 automated builds of different OpenWISP modules.
 
+``openwisp-utils-qa-checks``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can do multiple `checkmigrations` by passing the arguments with space-delimited string.
+
+For example, this multiple `checkmigrations`::
+
+    checkmigrations --migrations-to-ignore 3 \
+		    --migration-path ./openwisp_users/migrations/ || exit 1
+
+    checkmigrations --migrations-to-ignore 2 \
+		    --migration-path ./tests/testapp/migrations/ || exit 1
+
+Can be changed with::
+
+    openwisp-utils-qa-checks --migrations-to-ignore "3 2" --migration-path "./openwisp_users/migrations/ ./tests/testapp/migrations/"
+
 ``checkmigrations``
 ~~~~~~~~~~~~~~~~~~~
 
--- a/openwisp-utils-qa-checks
+++ b/openwisp-utils-qa-checks
@@ -18,6 +18,7 @@ show_help() {
   printf "  --migrations-to-ignore <int>\t: Number of migrations after which checking of migration file names should begin,\n"
   printf "\t\t\t\t  say, if checking needs to start after 0003_auto_20150410_3242.py value should be 3\n"
   printf "  --skip-checkmigrations\t: Skip migration name check\n"
+  printf "  You can do multiple migration check by passing the arguments with space-delimited string array.\n"
   printf "Commit name check options:\n"
   printf "  --message <commit-message>\t: Commit message (by default is equal to latest commit name)\n"
   printf "  --skip-checkcommit\t\t: Skip commit check\n"
@@ -52,15 +53,20 @@ runcheckendline() {
 }
 
 runcheckmigrations() {
-  echo "Running migration name check"
   if [ "$MIGRATION_PATH" == "" ]; then
     echoerr "ERROR: No migration path specified!"
     FAILURE=1
     return
   fi
-  checkmigrations --migration-path "$MIGRATION_PATH" --migrations-to-ignore "$MIGRATIONS_TO_IGNORE" \
-    && echo "SUCCESS: Migration name check successful!" \
-    || { echoerr "ERROR: Migration name check failed!"; FAILURE=1; }
+  for i in "${!MIGRATION_PATH[@]}"
+  do
+    path="${MIGRATION_PATH[$i]}"
+    ignore="${MIGRATIONS_TO_IGNORE[$i]}"
+    ignore="`[ $ignore ] && echo \"$ignore\" || echo 0`"
+    checkmigrations --migration-path "$path" --migrations-to-ignore "$ignore" \
+      && echo "SUCCESS: Migration name check on \"$path\" with migrations-to-ignore: $ignore successful!" \
+      || { echoerr "ERROR: Migration name check on \"$path\" with migrations-to-ignore: $ignore failed!"; FAILURE=1; }
+  done
 }
 
 runflake8() {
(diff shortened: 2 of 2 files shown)


marfgold1:
Wait, there is a little problem here. Not ready to review.

openwisp-utils-qa-checks
```diff
@@ -58,9 +59,16 @@ runcheckmigrations() {
...
     return
   fi
-  checkmigrations --migration-path "$MIGRATION_PATH" --migrations-to-ignore "$MIGRATIONS_TO_IGNORE" \
-    && echo "SUCCESS: Migration name check successful!" \
-    || { echoerr "ERROR: Migration name check failed!"; FAILURE=1; }
+  migrations_count="${#MIGRATION_PATH[@]}"
+  for i in "${!MIGRATION_PATH[@]}"
+  do
+    path="${MIGRATION_PATH[$i]}"
+    ignore="`[ ${MIGRATIONS_TO_IGNORE[$i]} ] && echo \"${MIGRATIONS_TO_IGNORE[$i]}\" || echo 0`"
```
atb00ker: Can you please maintain the output like the other tests ("SUCCESS", "ERROR"), like:
"SUCCESS: Migration name check on $path with migrations-to-ignore: $ignore"
"ERROR: Migration name check on $path with migrations-to-ignore: $ignore"
and remove the "Running.." line.

marfgold1: @atb00ker I did add something though before that like
```
Running migration name check...
[1/2] Checking on migration-path: $path with migrations-to-ignore: $ignore
<Traceback error>
ERROR: Migration name check failed!
[2/2] Checking on migration-path: <path> with migrations-to-ignore: <ignore>
<Traceback error>
ERROR: Migration name check failed!
```
It has a nice friendly section first then the traceback logs error instead of the one you propose that'll be like:
```
<Traceback error>
ERROR: Migration name check on $path with migrations-to-ignore: $ignore
<Traceback error>
ERROR: Migration name check on $path with migrations-to-ignore: $ignore
```
That write the traceback logs first then the error instead.
But if you really want to make it like the one you propose then okay I will change it, I just to make sure :smile:

atb00ker: Actually, I think 
```
<Traceback error>
ERROR: Migration name check on $path with migrations-to-ignore: $ignore
```
sounds good, even python does it that way and in most cases only. If it doesn't work out, we can always get back to this approach, what do you think? :smile:

marfgold1: :smile: Fair enough. Okay, I will change it

marfgold1:
@atb00ker Isn't it for another issue just like @nemesisdesign said in the original issue #52:
"We should also update the documentation accordingly when doing #47."

marfgold1:
Because no one doing that issue yet, then I can leave it to other that does #47 later.
Except if someone already did it and the PR that closes that issue is merged then I should

marfgold1:
Sorry I accidentally closed it :disappointed:

marfgold1:
@atb00ker done :smile: . I leave the description and other usage for that issue :sweat_smile:

README.rst
```diff
@@ -267,6 +267,23 @@ Quality Assurance checks
 This package contains some common QA checks that are used the
 automated builds of different OpenWISP modules.
 
+``openwisp-utils-qa-checks``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can do multiple `checkmigrations` by passing the arguments with space-delimited string.
+
+For example, this multiple `checkmigrations`::
```
nemesisdesign: There was a small mistake here, it needs to be ``checkmigrations`` with two slant quotes.

README.rst
```diff
@@ -267,6 +267,23 @@ Quality Assurance checks
 This package contains some common QA checks that are used the
 automated builds of different OpenWISP modules.
 
+``openwisp-utils-qa-checks``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can do multiple `checkmigrations` by passing the arguments with space-delimited string.
```
nemesisdesign: same here

marfgold1:
Whoops, sorry I don't know it should be two slant quotes for code, I thought it's only one just like github markdown. Also my bad I didn't check it carefully. Thanks for the help! @nemesisdesign 
And I see you make the `openwisp-utils-qa-checks` calls to be two line [here](https://github.com/openwisp/openwisp-utils/commit/2110cb8c821c2876418d13d3605cd6050c415892#diff-88b99bb28683bd5b7e3a204826ead112R285-R286), does we have rules for something like that or a documentation guidelines in general? Sorry if I missed something :disappointed_relieved:

PR #61, opened 2019-12-22
[tests] Increased text coverage to 100% #57

### LINKED ISSUE #57: [tests] Increase coverage to 100%
Increase test coverage to 100%.

oltarzewskik:
Fixes #57

**Depends on: https://github.com/openwisp/openwisp-utils/pull/62**

### DIFF
--- a/tests/test_project/tests/test_qa.py
+++ b/tests/test_project/tests/test_qa.py
@@ -1,4 +1,6 @@
+import io
 import os
+import sys
 from os import path
 
 from django.test import TestCase
@@ -55,10 +57,28 @@ class TestQa(TestCase):
             with patch('argparse._sys.argv', option):
                 try:
                     check_migration_name()
-                except (SystemExit, Exception):
+                except (SystemExit):
                     pass
                 else:
-                    self.fail('SystemExit or Exception not raised')
+                    self.fail('SystemExit not raised')
+
+    def test_migration_failure_message(self):
+        bad_migration = [
+            'checkmigrations',
+            '--migration-path', MIGRATIONS_DIR,
+        ]
+        with patch('argparse._sys.argv', bad_migration):
+            captured_output = io.StringIO()
+            sys.stdout = captured_output  # redirect stdout
+            try:
+                check_migration_name()
+            except (SystemExit):
+                message = 'must be renamed to something more descriptive'
+                self.assertIn(message, captured_output.getvalue())
+            else:
+                self.fail('SystemExit not raised')
+            finally:
+                sys.stdout = sys.__stdout__  # reset redirect
 
     def test_qa_call_check_commit_message_pass(self):
         options = [
@@ -184,10 +204,30 @@ class TestQa(TestCase):
             with patch('argparse._sys.argv', option):
                 try:
                     check_commit_message()
-                except (SystemExit, Exception):
+                except (SystemExit):
                     pass
                 else:
-                    self.fail('SystemExit or Exception not raised')
+                    self.fail('SystemExit not raised')
+
+    def test_commit_failure_message(self):
+        bad_commit = [
+            'commitcheck',
+            '--message',
+            "[qa] Updated file and fixed problem\n\n"
+            "Added more files. Fixes #20"
+        ]
+        with patch('argparse._sys.argv', bad_commit):
+            captured_output = io.StringIO()
+            sys.stdout = captured_output  # redirect stdout
+            try:
+                check_commit_message()
+            except (SystemExit):
+                message = 'Your commit message does not follow our commit message style guidelines'
+                self.assertIn(message, captured_output.getvalue())
+            else:
+                self.fail('SystemExit not raised')
+            finally:
+                sys.stdout = sys.__stdout__  # reset redirect
 
     def tearDown(self):
         os.unlink(self._test_migration_file)
--- a/openwisp_utils/admin_theme/admin.py
+++ b/openwisp_utils/admin_theme/admin.py
@@ -20,7 +20,7 @@ class OpenwispAdminSite(admin.AdminSite):
     )
 
 
-def openwisp_admin(site_url=None):
+def openwisp_admin(site_url=None):  # pragma: no cover
     """
     openwisp_admin function is deprecated and discouraged to use,
     you should use admin_site object from openwisp_utils.admin_theme.site instead
--- a/tests/test_project/admin.py
+++ b/tests/test_project/admin.py
@@ -1,9 +1,10 @@
 from django.contrib import admin
 from django.forms import ModelForm
-from openwisp_utils.admin import AlwaysHasChangedMixin, ReadOnlyAdmin
+from openwisp_utils.admin import (AlwaysHasChangedMixin, ReadOnlyAdmin,
+                                  TimeReadonlyAdminMixin)
 from openwisp_utils.admin_theme.site import admin_site
 
-from .models import Operator, Project, RadiusAccounting
+from .models import Operator, Project, RadiusAccounting, Shelf
 
 
 @admin.register(Operator, site=admin_site)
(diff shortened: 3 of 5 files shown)


tests/test_project/tests/test_qa.py
```diff
@@ -49,12 +50,19 @@ def test_qa_call_check_migration_name_failure(self):
...
+                'checkmigrations',
+                '--migration-path', MIGRATIONS_DIR,
+            ],
         ]
         for option in options:
             with patch('argparse._sys.argv', option):
                 try:
                     check_migration_name()
+                    message = 'must be renamed to something more descriptive'
+                    self.assertContains(io.StringIO().getvalue(),
```
atb00ker: Is this affecting the coverage?
Also, can this be moved to else clause?

tests/test_project/tests/test_qa.py
```diff
@@ -184,6 +198,9 @@ def test_qa_call_check_commit_message_failure(self):
             with patch('argparse._sys.argv', option):
                 try:
                     check_commit_message()
+                    message = 'must be renamed to something more descriptive'
+                    self.assertContains(io.StringIO().getvalue(),
+                                        message)
```
atb00ker: Umm.. why would commitcheck say "must be renamed to something more descriptive"? 
Also, the same questions as above? :smile:

oltarzewskik: @atb00ker It checks if failure message is printing.
tests these lines:
https://github.com/openwisp/openwisp-utils/blob/master/openwisp_utils/qa.py#L55
https://github.com/openwisp/openwisp-utils/blob/master/openwisp_utils/qa.py#L158

atb00ker: By "same questions as above, I meant:
- Is this affecting the coverage?
- Also, can this be moved to else clause?

atb00ker: Okay but read the message text. The commit text would not include "must be renamed to something more descriptive", right?

oltarzewskik: > Is this affecting the coverage?

yes it is, it checks two lines which weren't tested before

> Also, can this be moved to else clause?

yes, I can create 2 new functions instead

> Okay but read the message text. The commit error message would not include "must be renamed to something more descriptive", right?

according to https://github.com/openwisp/openwisp-utils/blob/master/openwisp_utils/qa.py#L33 it should, right?

oltarzewskik: > Also, can this be moved to else clause?

yes, I will move it in few hours

atb00ker: > according to https://github.com/openwisp/openwisp-utils/blob/master/openwisp_utils/qa.py#L33 it should, right?

but that's check migration name, I am highlighting the test called
"test_qa_call_check_commit_message_failure"
Why will commit_check test run migration_check, or am I missing something? :sweat_smile:

oltarzewskik: Oh, you are right, commit check should be totally different, fixed. thanks

oltarzewskik:
Before merging reverse all  changes in context_processor.py file because this [PR](https://github.com/openwisp/openwisp-utils/pull/62) fix that

tests/test_project/tests/test_qa.py
```diff
@@ -189,5 +207,23 @@ def test_qa_call_check_commit_message_failure(self):
...
+            "Added more files. Fixes #20"
+        ]
+        captured_output = io.StringIO()
+        sys.stdout = captured_output  # redirect stdout
+        with patch('argparse._sys.argv', bad_commit):
+            try:
+                check_commit_message()
+            except (SystemExit):
+                pass
+        sys.stdout = sys.__stdout__  # reset redirect
```
atb00ker: Do you need to do this even if you are inside the `patch`'s scope?

oltarzewskik: yes, even if I will move `sys.stdout = captured_output` to `patch` scope I will have to anyway reset `sys.stdout`

tests/test_project/tests/test_qa.py
```diff
@@ -189,5 +207,23 @@ def test_qa_call_check_commit_message_failure(self):
...
+            '--message',
+            "[qa] Updated file and fixed problem\n\n"
+            "Added more files. Fixes #20"
+        ]
+        captured_output = io.StringIO()
+        sys.stdout = captured_output  # redirect stdout
+        with patch('argparse._sys.argv', bad_commit):
+            try:
+                check_commit_message()
+            except (SystemExit):
```
atb00ker:
```
except (SystemExit):
   pass
```
makes me a little nervous, if something fails, one cannot tell the reason without editing the tests.

Can you please do it like:
```
                except (SystemExit, Exception) as e:
                    msg = 'Check failed:\n\n{}' \
                          '\n\nOutput:{}'.format(option[-1], e)
                    self.fail(msg)
                else:
                     sys.stdout = sys.__stdout__  # reset redirect
                     message = 'Your commit message does not follow our commit message style guidelines'
                     self.assertIn(message, captured_output.getvalue())
```

oltarzewskik: `SystemExit` in this test is supposed to be raised (look at these lines: https://github.com/openwisp/openwisp-utils/blob/master/openwisp_utils/qa.py#L157-#L159) and therefore if it will be raised everything in `else` clausule won't be executed.

We might supply user with message if `SystemExit` won't be raised (Exception or nothing would be raised), but wouldn't be better to do it in [this function](https://github.com/openwisp/openwisp-utils/blob/master/tests/test_project/tests/test_qa.py#L117)?

atb00ker: Hey, 

Ah, I haven't checked it that closely, I only wanted to point out the risk of 
```
except:
   pass
```
and we should avoid as much as we can unless it's the only clean way to test!

If it's suppose to execute the `except` part, even better, can we do:
```
except(SystemExit):
    sys.stdout = sys.__stdout__  # reset redirect
    message = 'Your commit message does not follow our commit message style guidelines'
    self.assertIn(message, captured_output.getvalue()
else: 
    self.fail("Error not raised")
```
After all, we only want to test in case the "except" occurs, right?
(Also, can something similar be done to avoid ```expect: pass``` in the other function as well?)

oltarzewskik: @atb00ker I think that even better would be something like this:
```
except(SystemExit):
    message = 'Your commit message does not follow our commit message style guidelines'
    self.assertIn(message, captured_output.getvalue()
else: 
    self.fail("Error not raised")
sys.stdout = sys.__stdout__  # reset redirect
```
because we want to reset redirect even if except won't occure
also, I don't understand why we except `Exception` [here](https://github.com/openwisp/openwisp-utils/blob/master/tests/test_project/tests/test_qa.py#L187) if we only except according to function `SystemExit`?

atb00ker: > because we want to reset redirect even if except won't occure

Ah, sorry, yep sounds good, if it needs to be executed in anycase it can be `finally:` or outside, both will work! :smile: 
The advantage of putting inside finally is that even if `except` fails for some reason, finally will still execute! :smile:  


> also, I don't understand why we except Exception here if we only except according to function SystemExit?

Even I don't know. Doesn't look good. @ppabcd can you help? :smile: :

ppabcd: > Even I don't know. Doesn't look good. @ppabcd can you help? 😄 :

Hmm, when I created this code, I saw the previous unit testing and copy it. After then, I modified to check for a commit message style check because I didn't know how to create a unit testing before.

oltarzewskik: >  I saw the previous unit testing and copy it

@ppabcd Can you send url to the "previous" unit test?

atb00ker: @oltarzewskik I think we can skip `Exception`, sometimes wrong code this gets merged. :smile:

PR #62, opened 2019-12-22
[feature] Added support for admin_site

oltarzewskik:
Added support for `admin_site` in `context_processor.py`

### DIFF
--- a/openwisp_utils/admin_theme/context_processor.py
+++ b/openwisp_utils/admin_theme/context_processor.py
@@ -1,7 +1,11 @@
+# TODO: remove once all openwisp modules are using admin_site
+import warnings
+
 from django.apps import registry
 from django.conf import settings
 from django.contrib.admin import site
 from django.urls import reverse
+from openwisp_utils.admin_theme.site import admin_site
 
 
 def menu_items(request):
@@ -29,7 +33,12 @@ def build_menu(request=None):
         url = reverse('admin:{}_{}_changelist'.format(app_label,
                                                       model.lower()))
         label = item.get('label', model_class._meta.verbose_name_plural)
-        model_admin = site._registry[model_class]
+        try:
+            model_admin = admin_site._registry[model_class]
+        except KeyError:
+            model_admin = site._registry[model_class]
+            warnings.warn('default django admin.site is deprecated by OpenWISP \
+                           please consider moving our to new admin_site')
         if not request or model_admin.has_module_permission(request):
             menu.append({
                 'url': url,


openwisp_utils/admin_theme/context_processor.py
```diff
@@ -3,6 +3,11 @@
 from django.contrib.admin import site
 from django.urls import reverse
 
+try:
+    from openwisp_utils.admin_theme.site import admin_site
+except Exception:  # pragma: no cover
```
nemesisdesign: I don't understand why are you doing this. Can you explain? It looks like something that shouldn't be done.
Why can it fail and why should it fail silently instead of loudly?

openwisp_utils/admin_theme/context_processor.py
```diff
@@ -29,7 +34,10 @@ def build_menu(request=None):
         url = reverse('admin:{}_{}_changelist'.format(app_label,
                                                       model.lower()))
         label = item.get('label', model_class._meta.verbose_name_plural)
-        model_admin = site._registry[model_class]
+        try:
+            model_admin = admin_site._registry[model_class]
+        except Exception:  # pragma: no cover
```
nemesisdesign: again, why use a generic exception? How can it fail?

I think we can register into both our admin_site and the default django one for backward compatibility and add a comment like:
```
# TODO: remove once all openwisp modules have been upgraded
```
above the line which is kept for backward compatibility.

openwisp_utils/admin_theme/context_processor.py
oltarzewskik: Oh, you are right there is no need for that, there is no way that it can fail

openwisp_utils/admin_theme/context_processor.py
oltarzewskik: @nemesisdesign Okay, I made some changes, please checkout

oltarzewskik:
@nemesisdesign Now basically all projects which will use the new `admin_site` and already use `context_processor` won't work, because now `context_processor` is checking only for `admin.site` registry therefore if will drop support for `admin.site` and add support `admin_site` in some project which use `context_processor`, `context_processor` will give `KeyError` like her: https://travis-ci.org/openwisp/openwisp-utils/jobs/629623077?utm_medium=notification&utm_source=github_status

I think that we can instead of merging two dictionaries we can do something like that:
```
try:
    model_admin = admin_site._registry[model_class]
except KeyError:
    model_admin = site._registry[model_class]
```

nemesisdesign:
@oltarzewskik please see https://github.com/openwisp/openwisp-utils/commit/1b9672134363bc00c28e720e29fa094e12c953c7

PR #71, opened 2020-01-11
[admin] Removed logic duplicates #40

### LINKED ISSUE #40: Generalize easy to copy uuid, secret key and secret URL logic

waleko:
_This PR is a follow-up of #41 by @NoumbissiValere_

Removed the logics which were duplicated
in the others modules and redefined it here to enable reusability and ease debugging.

Fixes #40


tests/test_project/tests/test_admin.py
```diff
@@ -52,17 +54,21 @@ def test_alwayshaschangedmixin(self):
...
             'name': 'test',
+            'key': self.TEST_KEY,
             'operator_set-TOTAL_FORMS': 1,
             'operator_set-INITIAL_FORMS': 1,
             'operator_set-MIN_NUM_FORMS': 0,
             'operator_set-MAX_NUM_FORMS': 1000,
             'operator_set-0-first_name': 'test2',
             'operator_set-0-last_name': 'test2',
-            'operator_set-0-id': project.pk,
+            'operator_set-0-id': operator.id,
```
waleko: There was a mistake here. It should be operator's id instead of projects. Previously, `project.pk` was just a number, which coincidentally was also equal to 1, but now `pk` is a uuid, therefore the test failed.

openwisp_utils/staticfiles.py
```diff
@@ -12,7 +12,7 @@ class DependencyFinder(FileSystemFinder):
     A static files finder that finds static files of
     django-apps listed in dependencies
     """
-    dependencies = EXTENDED_APPS
+    dependencies = EXTENDED_APPS + ['openwisp_utils']
```
waleko: It seems, that it's an easy fix to "putting `openwisp_utils` in `INSTALLED_APPS` problem"

openwisp_utils/admin.py
```diff
@@ -58,3 +60,64 @@ def has_changed(self):
...
+    def get_readonly_fields(self, request, obj=None):
+        return self._get_fields(self.readonly_fields, request, obj)
+
+    class Media:
+        js = ('admin/js/jquery.init.js', 'openwisp-utils/js/uuid.js')
+
+    uuid.short_description = _('UUID')
+
+
+class GetUrlAdmin(ModelAdmin):
```
waleko: I am also not sure if we need `GetUrlAdmin` as discussed [here](https://github.com/openwisp/openwisp-utils/issues/40#issuecomment-486012776). If we keep it, we have to come up with a test for it. Can someone help me to come up with the idea for the test then?

openwisp_utils/base.py
```diff
@@ -1,25 +1,50 @@
...
 
-class TimeStampedEditableModel(models.Model):
-    """
-    An abstract base class model that provides self-updating
-    ``created`` and ``modified`` fields.
-    """
-    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
-    created = AutoCreatedField(_('created'), editable=True)
-    modified = AutoLastModifiedField(_('modified'), editable=True)
+def _key_model_factory(key_name: str) -> Type[models.Model]:
```
waleko: I am not sure if this is a good solution to key problem. I made the factory, as in different repos the key is named differently. For example in [django-netjsonconfig it is `key`](https://github.com/openwisp/django-netjsonconfig/blob/master/django_netjsonconfig/base/device.py#L21-L26), but in [openwisp-radius it is called `token`](https://github.com/openwisp/openwisp-radius/blob/master/openwisp_radius/models.py#L142-L144)

openwisp_utils/validators.py
```diff
@@ -0,0 +1,8 @@
+from django.core.validators import RegexValidator, _lazy_re_compile
+from django.utils.translation import ugettext_lazy as _
+
+key_validator = RegexValidator(
+    _lazy_re_compile('^[^\\s/\\.]+$'),
+    message=_('Key must not contain spaces, dots or slashes.'),
```
nemesisdesign: Let's change the error message to `'This value must not contain spaces, dots or slashes.'

openwisp_utils/base.py
nemesisdesign: I think we're moving into too much abstraction territory with this one, you can remove it.

It's better to explicitly declare the key or token field each time for better readability and to make the code explicit therefore faster to inspect.

tests/test_project/tests/test_admin.py
nemesisdesign: good call

README.rst
```diff
@@ -34,6 +34,9 @@ Current features
 * **Customized admin theme** for OpenWISP modules
 * **TimeStamped** models and mixins which add self-updating ``created`` and ``modified`` fields.
 * **UUIDModel**: base model with a UUID4 primary key
+* **KeyModel**: base model with random unique string key
```
nemesisdesign: so we can remove this

README.rst
```diff
@@ -34,6 +34,9 @@ Current features
...
+* **UUIDAdmin**: base admin which defines a uuid field from a UUID primary key
+* **get_random_key**: generates an object key of 32 characters
```
nemesisdesign: we have to add a mention to `ReceiveUrlAdmin`

openwisp_utils/base.py
```diff
@@ -3,23 +3,38 @@
...
     class Meta:
         abstract = True
 
 
-class UUIDModel(models.Model):
-    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
-
-    class Meta:
-        abstract = True
+class KeyField(models.CharField):
```
waleko: @nemesisdesign I really don't want having to write `CharField` with all those parameters by hand every time. How about this custom field which is just a `CharField` with some default params?

tests/settings.py
```diff
@@ -16,6 +16,7 @@
     'django.contrib.staticfiles',
     'openwisp_utils.admin_theme',
     'django.contrib.sites',
+    'rest_framework',
```
waleko: @nemesisdesign This is for new api to work in the browser (with all those fancy drf html pages). This is not vitally important, we can remove this if you don't want it here.

waleko:
@nemesisdesign I fixed most of the stuff. 

I've [created `KeyField`](https://github.com/openwisp/openwisp-utils/pull/71/files#diff-215a8a7a1cebeee0027970ee5941566cR29-R40). It's just a CharField with some default parameters (e.g. validators, defaults). I really don't want having to write CharField with all those parameters by hand every time. 

I also modified `ReceiveUrlAdmin`. I wrote some tests for it. I may have gone for an overkill and [created a whole API for test_project](https://github.com/openwisp/openwisp-utils/tree/3951293b2cab31b30db5922d4801650d4780c371/tests/test_project/test_api) to test it.

openwisp_utils/admin.py
```diff
@@ -58,3 +62,79 @@ def has_changed(self):
...
+    Return a receive_url field whose value is that of
+    a view_name concatenated with the obj id and/or
+    with the key of the obj
+    """
+
+    receive_url_querystring_arg: str = 'key'
+    receive_url_object_arg: str = 'pk'
+    receive_url_name: str = None
+
+    def __init__(self, *args, **kwargs):
```
nemesisdesign: seems this method isn't doing anything and it can be removed

openwisp_utils/admin.py
```diff
@@ -58,3 +62,79 @@ def has_changed(self):
...
+    uuid.short_description = _('UUID')
+
+
+class ReceiveUrlAdmin(ModelAdmin):
+    """
+    Return a receive_url field whose value is that of
+    a view_name concatenated with the obj id and/or
+    with the key of the obj
+    """
+
```
nemesisdesign: remove this blank line please

tests/test_project/test_api/parsers.py
```diff
@@ -0,0 +1,21 @@
...
+    media_type = 'application/json'
+
+    def parse(self, stream, media_type=None, parser_context=None):
+        parser_context = parser_context or {}
+        encoding = parser_context.get('encoding', settings.DEFAULT_CHARSET)
+
+        try:
+            return stream.read().decode(encoding)
+        except ValidationError as e:  # pragma: nocover
+            raise ParseError('text/plain parse error - %s' % six.text_type(e))
```
nemesisdesign: this is definitely overkill

tests/test_project/test_api/views.py
```diff
@@ -0,0 +1,36 @@
...
+        # missing key: 400
+        if not key:
+            return Response({'detail': _('missing required "key" parameter')},
+                            status=400)
+        # wrong key 403
+        if project.key != key:
+            return Response({'detail': _('wrong key')},
+                            status=403)
+        return Response({'detail': _('ok'), 'name': project.name},
+                        status=200)
```
nemesisdesign: are all these if cases really necessary?

tests/test_project/admin.py
```diff
@@ -28,9 +29,16 @@ class OperatorInline(admin.StackedInline):
...
+class ProjectAdmin(UUIDAdmin, ReceiveUrlAdmin):
+    receive_url_name = 'receive_project'
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.inlines = [OperatorInline]
+        self.list_display = ('name',)
+        self.fields = ('uuid', 'name', 'key', 'receive_url',)
+        self.readonly_fields = ('uuid', 'receive_url',)
```
nemesisdesign: why you can't define these attributes as class attributes instead of having to define an `__init__` method for it?

tests/settings.py
nemesisdesign: why don't you just do it with a simple json view from django? If we add this here every project using openwisp-utils will have to download DRF which is definitely overkill

openwisp_utils/base.py
```diff
@@ -3,23 +3,38 @@
...
+    def __init__(self, *args, **kwargs):
+        fields = {
+            'max_length': 64,
+            'unique': True,
+            'db_index': True,
+            'default': get_random_key,
+            'validators': [key_validator],
+            'help_text': _('unique key')
```
nemesisdesign: remove this, we need to always explicitly define the `help_text`

openwisp_utils/base.py
```diff
@@ -3,23 +3,38 @@
...
-    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
-
-    class Meta:
-        abstract = True
+class KeyField(models.CharField):
+    def __init__(self, *args, **kwargs):
+        fields = {
+            'max_length': 64,
+            'unique': True,
+            'db_index': True,
```
nemesisdesign: same here, this needs to be defined explicitly

openwisp_utils/base.py
```diff
@@ -3,23 +3,38 @@
...
-class UUIDModel(models.Model):
-    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
-
-    class Meta:
-        abstract = True
+class KeyField(models.CharField):
+    def __init__(self, *args, **kwargs):
+        fields = {
+            'max_length': 64,
+            'unique': True,
```
nemesisdesign: same here, this needs to be defined explicitly

openwisp_utils/base.py
```diff
@@ -3,23 +3,38 @@
...
+            'db_index': True,
+            'default': get_random_key,
+            'validators': [key_validator],
+            'help_text': _('unique key')
+        }
+        fields.update(kwargs)
+        super(KeyField, self).__init__(*args, **fields)
```
nemesisdesign: use python 3 way of doing it, `super()` (rebase on the current master if you're still using the python 2 compatible version).

waleko:
@nemesisdesign Thanks for feedback.

I may have run into an issue here though. I'm currently implementing changes for django-netjsongraph. There is a [key field in `Topology` model](https://github.com/openwisp/django-netjsongraph/blob/master/django_netjsongraph/base/topology.py#L47-L53). I changed it to
```python
key = KeyField(unique=False,
               db_index=False,
               help_text=_('key needed to update topology from nodes '),
               verbose_name=_('key'),
               blank=True)
```
But now when I run `makemigrations` I get this error:
`TypeError: Couldn't reconstruct field key on django_netjsongraph.Topology: __init__() missing 2 required positional arguments: 'unique' and 'db_index'`.

Presumably, changing `CharField` to `KeyField` for old data somehow doesn't include `unique` and `db_index` arguments in the constructor of `KeyField`. You can see the whole commit in my commit [here](https://github.com/waleko/django-netjsongraph/commit/84b898aefa1f952e09e6283ee0ece2fe1a92ec60). I also tried to [fake the migrations file](https://github.com/waleko/django-netjsongraph/commit/84b898aefa1f952e09e6283ee0ece2fe1a92ec60#diff-905ed1d942a7436f620cdacbc1f46523) and then running `migrate`, but I get almost the same error: `TypeError: __init__() missing 2 required positional arguments: 'unique' and 'db_index'`.

The easy fix is of course to ditch `KeyField`, and just have `CharField`, but I would only do it as a last resort. Can you please help?

By the way, it `ReceiveUrlAdmin` works well for now.

openwisp_utils/base.py
```diff
@@ -3,23 +3,38 @@
...
 
-class UUIDModel(models.Model):
-    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
-
-    class Meta:
-        abstract = True
+class KeyField(models.CharField):
+    def __init__(self, unique: bool, db_index: bool, help_text: str, *args, **kwargs):
+        fields = {
+            'max_length': 64,
```
nemesisdesign: I think it's better to make this parameter overridable

waleko: @nemesisdesign Constructor accepts kwargs and then we update our fields dict with kwargs, so every parameter is overridable

waleko:
> @waleko open the PR on django-netjsongraph so I can see the code and test it

@nemesisdesign https://github.com/openwisp/django-netjsongraph/pull/109

openwisp_utils/admin.py
```diff
@@ -1,4 +1,8 @@
+from typing import Tuple, List
```
nemesisdesign: is it needed? My local `flake8` check is failing because of this line:

```
$ flake8
./openwisp_utils/admin.py:1:1: F401 'typing.Tuple' imported but unused
./openwisp_utils/admin.py:1:1: F401 'typing.List' imported but unused
```

openwisp_utils/base.py
```diff
@@ -3,23 +3,38 @@
...
+            'unique': unique,
+            'db_index': db_index,
+            'default': get_random_key,
+            'validators': [key_validator],
+            'help_text': help_text
+        }
+        fields.update(kwargs)
+        super().__init__(*args, **fields)
```
nemesisdesign: Please do the following:

move this field to a new `fields.py` file, since in `base.py` we are keeping the base models and is a good practice to separate logic in different files, plus `fields.py` is common across django packages.
After you do this you'll need to update your patch on django-netjsongraph.

Then update the code of the field as follows:

```python
class KeyField(models.CharField):
    def __init__(self,
                 max_length: int = 64,
                 unique: bool = False,
                 db_index: bool = False,
                 help_text: str = None,
                 *args, **kwargs):
        kwargs.update({
            'default': get_random_key,
            'validators': [key_validator],
        })
        super().__init__(max_length=max_length,
                         unique=unique,
                         db_index=db_index,
                         help_text=help_text,
                         *args, **kwargs)
```

Your version doesn't supply default values and hence initializing the field without an explicit `unique=False` will cause it to raise the exception.

Issue #72, opened 2020-01-17
[ci] Newline is not interpreted in openwisp-utils-qa-checks


dwang:
In the new `openwisp-utils-qa-checks`, there is a minor bug with the text printed.

```
openwisp-utils-qa-checks --skip-isort --skip-flake8 --skip-checkmigrations
Running blank endline check
SUCCESS: Blank endline check successful!
Running commit message check
SUCCESS: Commit name check successful!
File manage.py not found\nSkipping Make Migration Check
```

The last line does not have the newline interpreted correctly. The final result should look like 

```
openwisp-utils-qa-checks --skip-isort --skip-flake8 --skip-checkmigrations
Running blank endline check
SUCCESS: Blank endline check successful!
Running commit message check
SUCCESS: Commit name check successful!
File manage.py not found
Skipping Make Migration Check
```

PR #73, opened 2020-01-17
[ci] Newline is not interpreted in openwisp-utils-qa-checks #72

Weirdo914:
Changed echo to printf in pending migrations check

### DIFF
--- a/openwisp-utils-qa-checks
+++ b/openwisp-utils-qa-checks
@@ -98,7 +98,7 @@ runcheckcommit() {
 
 runcheckpendingmigrations() {
   if [ ! -f "./tests/manage.py" ]; then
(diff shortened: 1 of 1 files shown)


openwisp-utils-qa-checks
```diff
@@ -98,7 +98,7 @@ runcheckcommit() {
 
 runcheckpendingmigrations() {
   if [ ! -f "./tests/manage.py" ]; then
-    echo "File manage.py not found\nSkipping Make Migration Check"
```
nemesisdesign: we would still need a new line at the end of the string (`echo` automatically adds a new line at the end, `printf` doesn't).

What about simply use the following?

```shell
echo "File manage.py not found, skipping Make Migration Check."
```

PR #78, opened 2020-01-30
[enchancement] Optional urlconf for ReceiveURL

atb00ker:


README.rst
```diff
@@ -201,6 +201,14 @@ template ``context_processors`` in ``settings.py`` as shown below.
 Settings
 ^^^^^^^^
 
+``TOPOLOGY_API_URI``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
nemesisdesign: `~~~~~~~~~~~~~~~~~~~~~~~~~~~~~` should be if equal length to heading

README.rst
```diff
@@ -201,6 +201,14 @@ template ``context_processors`` in ``settings.py`` as shown below.
...
+
+**example**: ``http://topology.example.com/api/receive/``
```
nemesisdesign: Why aren't we defining the base url only?
If we do this we will have to define the URL for every similar use cases.
Suppose we add another API that we call in the admin, we'll have to add a new setting for each case.

Can you change the logic so that only the base url is passed and the rest of the URL path is still calculated by the code?

The setting could be named ``TOPOLOGY_API_URL``, I would specify that it defaults to `None` and it's automatically calculated.

atb00ker: I see you point. Yes, adding a variable for every URL sounds bad. 
However, if the user doesn't mention `/api/receive/`, I'll have to hard code it, that sucks too!

atb00ker: -- Made the changes --

openwisp_utils/admin.py
```diff
@@ -108,26 +111,41 @@ class ReceiveUrlAdmin(ModelAdmin):
...
         if self.receive_url_name is None:
             raise ValueError('receive_url_name is not set up')
-        reverse_kwargs = {}
-        if self.receive_url_object_arg:
-            reverse_kwargs = {
-                self.receive_url_object_arg: getattr(obj, self.receive_url_object_arg)
-            }
-        url = reverse(self.receive_url_name, kwargs=reverse_kwargs)
+        if hasattr(settings, "TOPOLOGY_API_URL"):
+            url = settings.TOPOLOGY_API_URL + '/api/receive/'
```
nemesisdesign: ok sorry for not being more explicit.
This is a base module, it must not know anything about any other higher level openwisp module, otherwise we would be creating a mess of interdependency.

The URL that needs to be reversed is configurable as a class attribute right?
So we need to add a new class attribute (eg: `base_url`) defaulting to `None` to `ReceiveUrlAdmin` and use it in its code, that is in django-netjsongraph we must add the base_url class attribute taken from a setting like you do here, I'll try to provide some more examples in the next comments

openwisp_utils/admin.py
```diff
@@ -108,26 +111,41 @@ class ReceiveUrlAdmin(ModelAdmin):
...
         :param obj: Object for which the url is generated
         """
         if self.receive_url_name is None:
             raise ValueError('receive_url_name is not set up')
-        reverse_kwargs = {}
-        if self.receive_url_object_arg:
-            reverse_kwargs = {
-                self.receive_url_object_arg: getattr(obj, self.receive_url_object_arg)
-            }
-        url = reverse(self.receive_url_name, kwargs=reverse_kwargs)
```
nemesisdesign: this code was ok, `reverse` would generate something like `/api/url`.

If `self.base_url` is `None`, nothing would change, if `self.base_url` is not `None`, it would be prepended to it.

Then here: https://github.com/openwisp/django-netjsongraph/blob/master/django_netjsongraph/base/admin.py#L48
You would add the setting you added here, the setting would default to `None` and in docker-openwisp it will default to the topology module URL, is it clearer now?

atb00ker: Yes, this sounds like a much better approach than hard-coding anything! :smile:

However, consider the usage in a server where topology api is not registered because topology server is different, like in docker-openwisp:

1. User inherits `TopologyAdmin` in it's admin & changes the value of `base_url` to the correct `base_url`.
2 User adds a fake `urlpattern` (say `/api/receive`) with name `receive_url_name` in the `dashboard` container. (fake because `dashboard.openwisp.org/api/receive` doesn't point to anything ) but the `reverse` will be used to create the link `topology.server.com/api/receive`

The second point is why I didn't implement it like that in the first place. It feels like a hack to create a link that doesn't point to anything.
Is this what you have in mind or do I still have error in my thinking?

nemesisdesign: eg:

```python
url = reverse(self.receive_url_name, urlconf=self.receive_url_urlconf, kwargs=reverse_kwargs)
``

We then need another attribute eg: `receive_url_baseurl`

atb00ker: Made the changes

PR #89, opened 2020-04-06
[black] Add black support

### LINKED ISSUE #86: Improve QA checks: black, re-format code, commit message check, explain in docs
We've been spending a lot of time on code formatting during reviews lately.

In the past we discussed introducing [black](https://github.com/psf/black), I think we could start introducing it, I leave some notes here:

- add it to the QA requirements
- add it to the [QA checks](https://github.com/openwisp/openwisp-utils/blob/master/openwisp-utils-qa-checks) (ensure there's a way to skip the check)
- add a way to trigger the reformatting using isort and black (eg: `openwisp-qa-format`), so we can format the code with one command (hence making it easy for new contributors)
- update the python code convetions docs to suggest the use of `openwisp-qa-format` and `openwisp-qa-check`: http://openwisp.io/docs/developer/contributing.html#python-code-conventions
- add a flag to force the commit message check and add the suggestion of using this feature in the guidelines at: http://openwisp.io/docs/developer/contributing.html#commit-message-style-guidelines

PabloCastellano:
This allows us to define a common code formatting standard and avoids
having to manually do these changes manually.
It is enabled by default and can be disabled using --skip-black
This PR also formats the code in this repository using black.
Also added flag --force-checkcommit to force the commit message check

Fixes #86

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -58,7 +58,7 @@ Install from pypi:
 .. code-block:: shell
 
     pip install openwisp-utils
-    # install optional dependencies for tests (flake8 and isort)
+    # install optional dependencies for tests (flake8, black and isort)
     pip install openwisp-utils[qa]
 
 Install development version
@@ -327,7 +327,7 @@ This package contains some utilities that are used in the automated builds
 of different OpenWISP modules.
 
 ``catch_signal``
-~~~~~~~~~~~~~~~~ 
+~~~~~~~~~~~~~~~~
 
 This method can be used to mock a signal call inorder to easily verify
 that the signal has been called.
@@ -350,6 +350,11 @@ Quality Assurance checks
 This package contains some common QA checks that are used in the
 automated builds of different OpenWISP modules.
 
+``openwisp-qa-format``
+~~~~~~~~~~~~~~~~~~~~~~
+
+Shell script to automatically format Python code. It runs ``isort`` and ``black``.
+
 ``openwisp-utils-qa-checks``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
@@ -361,6 +366,7 @@ Shell script to run the following quality assurance checks:
 * `checkpendingmigrations <#checkpendingmigrations>`_
 * ``flake8`` - Python code linter
 * ``isort`` - Sorts python imports alphabetically, and seperated into sections
+* ``black`` - Formats python code using a common standard
 
 If a check requires a flag, it can be passed forward in the same way.
 
--- a/openwisp-qa-format
+++ b/openwisp-qa-format
@@ -1,0 +1,5 @@
+#!/bin/bash
+set -e
+
+isort -y . --recursive
+black -S .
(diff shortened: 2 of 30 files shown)


openwisp_utils/__init__.py
```diff
@@ -1,18 +1,18 @@
-VERSION = (0, 4, 4, 'final')
+VERSION = (0, 4, 4, "final")
```
atb00ker, line 1: Does black recommend using double quotes? 
( If it does, that sucks, I just converted double quotes to single quotes in controller at a lot of places! :joy: )

PabloCastellano: This is the default. We can configure it to use single quotes :)

openwisp-utils-qa-checks
```diff
@@ -82,6 +85,12 @@ runisort() {
     || { echoerr "ERROR: Isort check failed!"; FAILURE=1; }
 }
 
+runblack() {
+  black --check --diff . \
```
nemesisdesign, line 89: add the `-S` flag because most modules use mainly single quotes, so we can avoid having to reformat all the files of each module which will help us when reviewing historic changes in the future

nemesisdesign: @atb00ker :point_up: :blush:

openwisp-utils-qa-checks
```diff
@@ -82,6 +85,12 @@ runisort() {
...
+    && echo "SUCCESS: Black check successful!" \
+    || { echoerr "ERROR: Black check failed!"; FAILURE=1; }
```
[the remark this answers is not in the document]
PabloCastellano: done

openwisp-utils-qa-checks
PabloCastellano: done

openwisp-qa-format
```diff
@@ -0,0 +1,23 @@
...
+    || { echoerr "ERROR: Black run failed!"; FAILURE=1; }
+}
+
+
+set -e
+
+runisort
+runblack
+
+exit $FAILURE
```
nemesisdesign, line 23: I think this script can be simplified to:

```python
#!/bin/bash
set -e

isort -y . --recursive
black -S .
```

The -y option on isort is used to avoid asking for confirmation. We just run it.

If anything goes wrong with isort and black, those programs will provide the necessary info and exit code, but it should not happen, isort -y never failed since I started using it, so we can avoid worrying about catching failures in this case and keep the script super simple.

README.rst
```diff
@@ -350,6 +350,11 @@ Quality Assurance checks
 This package contains some common QA checks that are used in the
 automated builds of different OpenWISP modules.
 
+``openwisp-qa-format``
+~~~~~~~~~~~~~~~~~~~~~~
+
+Shell script to automatically format Python code. It runs `isort` and `black`.
```
nemesisdesign, line 356: in ReStructuredText you need to use two slant quotes:

```
It runs ``sort`` and ``black``.
```

PR #90, opened 2020-04-06
[checkcommit] Add argument to force message commit check #86

### LINKED ISSUE #86: Improve QA checks: black, re-format code, commit message check, explain in docs

PabloCastellano:
Previously this check was only run in Pull Requests.
Now it is possible to force it by using --force-checkcommit

Partially fixes #86

### DIFF
--- a/openwisp-utils-qa-checks
+++ b/openwisp-utils-qa-checks
@@ -4,7 +4,7 @@ show_help() {
   printf "Usage:\n"
   printf "  openwisp-utils-qa-checks --help\n"
   printf "  openwisp-utils-qa-checks {--migration-path <path> [--migrations-to-ignore <int>] | --skip-checkmigrations}\n"
-  printf "            {[--message <commit-message>] | --skip-checkcommit}\n"
+  printf "            {[--message <commit-message>] | --skip-checkcommit | --force-checkcommit}\n"
   printf "            {[--migration-module <module>] | --skip-checkmakemigrations}\n"
   printf "            [--skip-checkendline]\n"
   printf "            [--skip-flake8]\n"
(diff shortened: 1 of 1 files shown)


PabloCastellano:
@nemesisdesign Commit message changed :+1: 

Now thinking about this PR, having --force-checkcommit is not useful at all. Having changed the condition to run the check (`if [ "$SKIP_CHECKCOMMIT" = "false" -o "$TRAVIS_PULL_REQUEST" != "false" ]; then`) and having SKIP_CHECKCOMMIT=false as default, makes the --force argument useless

nemesisdesign:
@PabloCastellano Just found out now. Can you fix that in order to maintain the desired behaviour?

The check should be done automatically only in pull requests, it should be forced only with the flag.
The skip flag is needed to allow adding a `#noqa` flag to the commit to skip the check.

PR #93, opened 2020-04-06
[checkcommit] Fix logic for --skip-checkcommit and --force-checkcommit #86

PabloCastellano:
This check is run when any of the following conditions is true:

1. --force-checkcommit is specified
2. This function is invoked from Travis building a Pull Request

Context: https://github.com/openwisp/openwisp-utils/pull/90#issuecomment-609922762

Related to #86

### DIFF
--- a/openwisp-utils-qa-checks
+++ b/openwisp-utils-qa-checks
@@ -4,7 +4,7 @@ show_help() {
   printf "Usage:\n"
   printf "  openwisp-utils-qa-checks --help\n"
   printf "  openwisp-utils-qa-checks {--migration-path <path> [--migrations-to-ignore <int>] | --skip-checkmigrations}\n"
-  printf "            {[--message <commit-message>] | --skip-checkcommit}\n"
+  printf "            {[--message <commit-message>] | --skip-checkcommit | --force-checkcommit}\n"
   printf "            {[--migration-module <module>] | --skip-checkmakemigrations}\n"
   printf "            [--skip-checkendline]\n"
   printf "            [--skip-flake8]\n"
@@ -25,6 +25,7 @@ show_help() {
   printf "  --migration-module <module>\t: Name of Migration Module\n"
   printf "  --skip-checkmakemigrations\t: Skip make migration check\n"
   printf "  --skip-checkcommit\t\t: Skip commit check\n"
+  printf "  --force-checkcommit\t\t: Force commit check\n"
   printf "Blank endlines check options:\n"
   printf "  --skip-checkendline\t\t: Skip blank endlines check\n"
   printf "Flake8 check options:\n"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -22,7 +22,7 @@ install:
   - pip install -U -r requirements-test.txt
 
 before_script:
-  - openwisp-utils-qa-checks --migration-path "./tests/test_project/migrations"
+  - openwisp-utils-qa-checks --force-checkcommit --migration-path "./tests/test_project/migrations"
```
nemesisdesign, line 25: let's not do this otherwise this build will fail at every merge

openwisp-utils-qa-checks
```diff
@@ -83,15 +83,18 @@ runisort() {
     || { echoerr "ERROR: Isort check failed!"; FAILURE=1; }
 }
 
+# This check is run when any of the following conditions is true:
+# 1. --force-checkcommit is specified
+# 2. This function is invoked from Travis building a Pull Request
 runcheckcommit() {
-  [ -n "$TRAVIS_PULL_REQUEST" ] || TRAVIS_PULL_REQUEST="false"
-  if [ "$SKIP_CHECKCOMMIT" = "false" -o "$TRAVIS_PULL_REQUEST" != "false" ]; then
+	if [[ $FORCE_CHECKCOMMIT = true ]] ||
```
nemesisdesign, line 90: indentation is odd, can you put this at the same indentation level as it was before please?

openwisp-utils-qa-checks
```diff
@@ -83,15 +83,18 @@ runisort() {
...
+		 [[ $TRAVIS = true && $TRAVIS_PULL_REQUEST = true ]]; then
```
nemesisdesign, line 91: what is `$TRAVIS = true`? Sounds like something that can break existing builds, please remove it.
We want to run this check only on pull requests for now.

openwisp-utils-qa-checks
```diff
@@ -83,15 +83,18 @@ runisort() {
...
     echo "Running commit message check"
     checkcommit --message "$COMMIT_MESSAGE" \
     && echo "SUCCESS: Commit name check successful!" \
     || { printf "Checking commit message:\n\n"; printf "%s\n\n" "$COMMIT_MESSAGE"; echoerr "ERROR: Commit name check failed!"; FAILURE=1; }
   else
-    echo "SUCCESS: Commit name check successful!"
+    echo "SKIPPED: Commit name check"
```
nemesisdesign, line 97: spotted something odd in the old code, can you replace the instances of "Commit name" with "Commit message" please?

openwisp-utils-qa-checks
```diff
@@ -83,15 +83,18 @@ runisort() {
     || { echoerr "ERROR: Isort check failed!"; FAILURE=1; }
 }
 
+# This check is run when any of the following conditions is true:
+# 1. --force-checkcommit is specified
+# 2. This function is invoked from Travis building a Pull Request
 runcheckcommit() {
-  [ -n "$TRAVIS_PULL_REQUEST" ] || TRAVIS_PULL_REQUEST="false"
```
nemesisdesign, line 87: I think this line was fine, I would not remove it

openwisp-utils-qa-checks
PabloCastellano: `TRAVIS = true` when the script is being invoked by Travis (https://docs.travis-ci.com/user/environment-variables/#default-environment-variables) but you are right. Removed.

PR #97, opened 2020-04-08
[checks] Rename openwisp-utils-qa-checks to openwisp-qa-check #94

### LINKED ISSUE #94: [qa] Rename openwisp-utils-qa-checks to openwisp-qa-check
For consistency with the new `openwisp-qa-format`, let's rename `openwisp-utils-qa-checks` to `openwisp-qa-check` but we have to maintain backward compatibility with the old command (`openwisp-utils-qa-checks`) to avoid breaking builds.

PabloCastellano:
The old script name invokes the new one and adds a scary deprecation warning.
This way we are keeping backwards compatibility until we decide to remove
it in the future.

Fixes #94


PabloCastellano:
Done

PR #104, opened 2020-05-09
[admin-theme] Fix permission evaluation when building the menu

### LINKED ISSUE #103: [admin-theme] menu doesn't respect assigned permissions
The links "Templates" and "VPN Servers" are available in the menu although the user doesn't have the rights for these objects:

grafik

grafik

I had a look into this a bit:

https://github.com/openwisp/openwisp-utils/blob/master/openwisp_utils/admin_theme/context_processor.py#L32

* `ModelAdmin.has_module_permission` evaluates permissions only on the module level, not on the model level (Device, Template, Vpn).
* `openwisp-utils` is decoupled from the other OpenWISP modules so we don't have access to models etc

okraits:
Fixes #103

### DIFF
--- a/openwisp_utils/admin_theme/context_processor.py
+++ b/openwisp_utils/admin_theme/context_processor.py
@@ -1,6 +1,5 @@
 from django.apps import registry
 from django.conf import settings
-from django.contrib.admin import site
 from django.urls import reverse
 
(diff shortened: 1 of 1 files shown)


nemesisdesign:
Thanks @okraits, merged in #123

PR #105, opened 2020-05-09
[rest] Add optional swagger API endpoints #102

### LINKED ISSUE #102: [rest] Add optional swagger API endpoints
- add `extra_requires['rest']` with DRF and swagger in setup.py
- add the URLs for enabling the swagger docs, these URLs will be included in the test project of the other modules
- add a setting named `OPENWISP_REST_SWAGGER` (or similar) which shall control whether the swagger docs is enabled or not (some users in prod may want to disable this)
- ensure the test coverage does not decrease
- add basic info about this feature in the README

Related to #91.

PabloCastellano:
Closes #102

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -312,6 +312,24 @@ Example usage:
         '/static/custom-admin-theme.js',
     ]
 
+``OPENWISP_REST_SWAGGER``
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**default**: ``True``
+
+Enable or disable Swagger API documentation endpoint.
+
+When enabled, you can view the available documentation using the
+Swagger endpoint at ``/api/v1/swagger/``.
+
+You also need to add the following url to your project urls.py:
+
+.. code-block:: python
+
+    urlpatterns += [
+        url(r'^api/v1/', include('openwisp_utils.api.urls')),
+    ]
+
 Admin mixins
 ^^^^^^^^^^^^
 
--- a/README.rst
+++ b/README.rst
@@ -330,6 +330,26 @@ You also need to add the following url to your project urls.py:
         url(r'^api/v1/', include('openwisp_utils.api.urls')),
     ]
 
+``OPENWISP_API_INFO``
+~~~~~~~~~~~~~~~~~~~~~
+
+**default**:
+
+.. code-block:: python
+
+    {
+        'title': 'OpenWISP API',
+        'default_version': 'v1',
+        'description': 'OpenWISP REST API',
+    }
+
+Define OpenAPI general information.
+NOTE: This setting requires ``OPENWISP_REST_SWAGGER = True`` to take effect.
+
+For more information about optional parameters check the
+`drf-yasg documentation <https://drf-yasg.readthedocs.io/en/stable/readme.html#quickstart>`_
+.
+
 Admin mixins
 ^^^^^^^^^^^^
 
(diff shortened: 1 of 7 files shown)


tests/settings.py
```diff
@@ -85,6 +86,8 @@
 # during development only
 EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
 
+OPENWISP_REST_SWAGGER = True
```
nemesisdesign, line 89: Is this necessary? The setting defaults to True so this here is redundant.

tests/settings.py
```diff
@@ -16,6 +16,7 @@
     'django.contrib.staticfiles',
     'openwisp_utils.admin_theme',
     'django.contrib.sites',
+    'openwisp_utils',
```
nemesisdesign, line 19: is this really necessary? It should not be right?

Isn't drf-yasg needed? If it is, don't forget to mention it in the README section you're adding.

openwisp_utils/api/urls.py
```diff
@@ -0,0 +1,38 @@
...
+
+urlpatterns = []
+
+if app_settings.REST_SWAGGER:
+    schema_view = get_schema_view(
+        openapi.Info(
+            title='OpenWISP API',
+            default_version='v1',
+            description='OpenWISP REST API',
+            contact=openapi.Contact(email='openwisp@googlegroups.com'),
```
nemesisdesign, line 15: can we omit this? That email address does not work.

README.rst
```diff
@@ -312,6 +312,13 @@ Example usage:
         '/static/custom-admin-theme.js',
     ]
 
+``OPENWISP_REST_SWAGGER``
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**default**: ``True``
+
+Enable or disable Swagger API documentation endpoint.
```
nemesisdesign, line 320: I would specify how to access the swagger API and that to enable it is required to include the urls in the root urlconf, eg: ` url(r'^api/v1/', include('openwisp_utils.api.urls'))`

tests/settings.py
PabloCastellano: oh, right

tests/settings.py
PabloCastellano: I'm confused, do you know why adding the line to urls.py fails?

https://travis-ci.org/github/openwisp/openwisp-utils/jobs/685020239#L405

nemesisdesign: In my experience, that may be indicating another import error is being detected in that file.

I would create a new virtualenv and execute all the commands in `.travis.yml` (exactly in the order used by travis) to replicate its behavior and find out if the error is replicated.

PabloCastellano: Looks like the problems was that we need to run `pip install -e .`... as the last command otherwise openwisp-utils from pip is installed which does not contains the urls.py file

openwisp_utils/api/urls.py
```diff
@@ -0,0 +1,38 @@
...
+            license=openapi.License(name="GPLv3"),
```
nemesisdesign, line 16: I'm pretty sure users will need to modify this info since they tend to use openwisp as a base for their tailored solutions.

could you make all these arguments configurable with a dict?

Eg:

```python
API_INFO = getattr(settings, 'OPENWISP_API_INFO', {
    # ...
})

openapi.Info(**API_INFO)
```

PabloCastellano: done

PR #112, opened 2020-05-20
[api] Implement ScopedRateThrottle by default

PabloCastellano:
This PR is part of https://github.com/openwisp/openwisp-firmware-upgrader/issues/48

Note that by using ScopedRateThrottling it requires us to define a default anon class throttle for the unauthenticated API requests.

### DIFF
--- a/tests/test_project/apps.py
+++ b/tests/test_project/apps.py
@@ -1,18 +1,14 @@
 from django.apps import AppConfig
 from django.conf import settings
 
-from openwisp_utils import settings as app_settings
-
 
 class TestAppConfig(AppConfig):
     name = 'test_project'
     label = 'test_project'
 
-    _default_drf = {
+    DEFAULT_REST_FRAMEWORK = {
         'DEFAULT_THROTTLE_CLASSES': ['rest_framework.throttling.ScopedRateThrottle'],
-        'DEFAULT_THROTTLE_RATES': {
-            'anon': app_settings.API_RATE_ANON,
-        },
+        'DEFAULT_THROTTLE_RATES': {'anon': '40/hour'},
     }
 
     def ready(self, *args, **kwargs):
@@ -29,5 +25,6 @@ class TestAppConfig(AppConfig):
 
     def configure_drf_defaults(self):
         config = getattr(settings, 'REST_FRAMEWORK', {})
-        config.update(self._default_drf)
+        for key in self.DEFAULT_REST_FRAMEWORK.keys():
+            config.setdefault(key, self.DEFAULT_REST_FRAMEWORK[key])
         setattr(settings, 'REST_FRAMEWORK', config)
--- a/tests/test_project/apps.py
+++ b/tests/test_project/apps.py
@@ -1,14 +1,24 @@
 from django.apps import AppConfig
 from django.conf import settings
 
+from openwisp_utils import settings as app_settings
+
 
 class TestAppConfig(AppConfig):
     name = 'test_project'
     label = 'test_project'
 
+    _default_drf = {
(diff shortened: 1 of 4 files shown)


tests/test_project/apps.py
```diff
@@ -1,18 +1,33 @@
...
         menu_setting = 'OPENWISP_DEFAULT_ADMIN_MENU_ITEMS'
         items = [
             {'model': 'test_project.Shelf'},
         ]
         setattr(settings, menu_setting, items)
+
+    def configure_drf_defaults(self):
+        config = getattr(settings, 'REST_FRAMEWORK', {})
+        config.update(self._default_drf)
+        setattr(settings, 'REST_FRAMEWORK', config)
```
nemesisdesign, line 33: It looks to me that this way of doing it does not allow users to override our defaults.

I tried putting this in test/settings.py:

REST_FRAMEWORK = {'DEFAULT_THROTTLE_RATES': {'anon': '100/hour'}}

But in the django shell I see `40/hour:

$ ./manage.py shell
In [1]: from django.conf import settings                                                                                                                                                                           

In [2]: settings.REST_FRAMEWORK                                                                                                                                                                                    
Out[2]: 
{'DEFAULT_THROTTLE_CLASSES': ['rest_framework.throttling.ScopedRateThrottle'],
 'DEFAULT_THROTTLE_RATES': {'anon': '40/hour'}}

I think it would be better to allow users to override our defaults, we could do something like the following:

- does the project settings file define `REST_FRAMEWORK['DEFAULT_THROTTLE_CLASSES']`? If yes, don't override it
- before setting `self._default_drf['DEFAULT_THROTTLE_RATES']` check if each key is already set in `REST_FRAMEWORK['DEFAULT_THROTTLE_RATES']`, if it is, skip

With this approach we would be able to:

- supply our own default values
- allow users to override them
- avoid the need of defining settings like `OPENWISP_API_RATE_ANON`, because they can just do:

```python
REST_FRAMEWORK = {'DEFAULT_THROTTLE_RATES': { 'anon': '100/hour'}}
```

This way of doing this is also familiar to many django developers and is also [documented in the DRF docs](https://www.django-rest-framework.org/api-guide/throttling/#setting-the-throttling-policy), which we could link from our own docs, so we would be following a standard that is already well known in the community.

What do you think?

tests/test_project/apps.py
```diff
@@ -1,18 +1,33 @@
...
+        self.configure_drf_defaults()
 
     def add_default_menu_items(self):
         menu_setting = 'OPENWISP_DEFAULT_ADMIN_MENU_ITEMS'
         items = [
             {'model': 'test_project.Shelf'},
         ]
         setattr(settings, menu_setting, items)
+
+    def configure_drf_defaults(self):
```
nemesisdesign, line 30: I'm also thinking that this method could be part of a mixin which we could import in the other modules so we don't need to redefine it each time, we will just add the mixin to the AppConfig class, or alternatively we can define a function and call that function in the modules (instead of using the mixin, which some people don't like, I'm fine with both approaches).

tests/test_project/apps.py
```diff
@@ -1,18 +1,33 @@
...
 from django.conf import settings
 
+from openwisp_utils import settings as app_settings
+
 
 class TestAppConfig(AppConfig):
     name = 'test_project'
     label = 'test_project'
 
+    _default_drf = {
```
nemesisdesign, line 11: what about a more explicit class attribute like `DEFAULT_REST_FRAMEWORK_SETTINGS`?

tests/test_project/apps.py
PabloCastellano: Can't agree more!

PR #113, opened 2020-05-20
[utils] Added deep_merge_dicts

pandafy:
See https://github.com/openwisp/openwisp-notifications/pull/19#discussion_r428313884

### DIFF
--- a/openwisp_utils/utils.py
+++ b/openwisp_utils/utils.py
@@ -6,3 +6,14 @@ def get_random_key():
     generates a device key of 32 characters
     """
     return get_random_string(length=32)
+
+
+def deep_merge_dicts(dict1, dict2):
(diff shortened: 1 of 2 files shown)


atb00ker:
@pandafy it would be nice to know what this change does, could you please share where can this function be useful? :smile:

pandafy:
The function recursively merges two dictionaries. It was getting used in `openwisp-notifications` (but not any more) and [openwisp-monitoring](https://github.com/openwisp/openwisp-monitoring/blob/ffca6593d24689f2e77258a49a9361e09c294361/openwisp_monitoring/monitoring/charts.py#L137). It is helpful for updating default configurations defined using `dict`. 

The idea was to remove repetition across OpenWISP modules. :smile:

PR #125, opened 2020-07-31
[fix] Add support for docstrings in TimeLoggingTestRunner #124

### LINKED ISSUE #124: [bug] Timelogging test runner breaks down if docstring is used in a slow test
I was testing some changes in a test in `openwisp_monitoring` and thought of using a docstring to better explain a test.
That's when I noticed `TimeLoggingTestRunner` doesn't work with `docstrings`, due to one particular line. Below is the traceback, it can be easily reproduced by introducing a docstring in any slow test.

```python
Traceback (most recent call last):
  File "tests/test_project/tests/test_test_utils.py", line 51, in test_time_logging_runner
    runner.run_suite(suite)
  File "/home/hardik/.local/lib/python3.6/site-packages/django/test/runner.py", line 630, in run_suite
    return runner.run(suite)
  File "/usr/lib/python3.6/unittest/runner.py", line 180, in run
    stopTestRun()
  File "/home/hardik/openwisp/openwisp-utils/openwisp_utils/tests.py", line 66, in stopTestRun
    self.display_slow_tests()
  File "/home/hardik/openwisp/openwisp-utils/openwisp_utils/tests.py", line 51, in display_slow_tests
    name, module = name.split()
ValueError: too many values to unpack (expected 2)
```
Though I haven't seen much usage of docstring in tests in `openwisp`, still there shouldn't be a limitation, in my opinion. I will try to fix this and submit a patch.

nepython:
Closes #124

The actual fix is only one line,
```python
# Remove docstring if present
name = name.split('\n')[0]
```
Rest all restructuring was needed for the test, since the bug will show up only for a slow test with a docstring.

### DIFF
--- a/openwisp_utils/tests.py
+++ b/openwisp_utils/tests.py
@@ -20,13 +20,18 @@ def catch_signal(signal):
 
 
 class TimeLoggingTestResult(TextTestResult):
-    slow_test_threshold = getattr(settings, 'OPENWISP_SLOW_TEST_THRESHOLD', [0.3, 1])
-    assert isinstance(slow_test_threshold, list)
-    assert len(slow_test_threshold) == 2
-
     def __init__(self, *args, **kwargs):
         self.test_timings = []
+        self.slow_test_threshold = self._get_slow_test_threshold()
         super().__init__(*args, **kwargs)
+
+    def _get_slow_test_threshold(self):
+        slow_test_threshold = getattr(
+            settings, 'OPENWISP_SLOW_TEST_THRESHOLD', [0.3, 1]
+        )
+        assert isinstance(slow_test_threshold, list)
+        assert len(slow_test_threshold) == 2
+        return slow_test_threshold
 
     def startTest(self, test):
         self._start_time = time()
@@ -48,6 +53,8 @@ class TimeLoggingTestResult(TextTestResult):
         for name, elapsed in self.test_timings:
             if elapsed > self.slow_test_threshold[0]:
                 slow_tests_counter += 1
+                # Remove docstring if present
+                name = name.split('\n')[0]
                 name, module = name.split()
                 if module != self._module:
                     self._module = module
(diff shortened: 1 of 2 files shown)


openwisp_utils/tests.py
```diff
@@ -48,6 +53,8 @@ def display_slow_tests(self):
         for name, elapsed in self.test_timings:
             if elapsed > self.slow_test_threshold[0]:
                 slow_tests_counter += 1
+                # Remove docstring if present
+                name = name.split('\n')[0]
```
nemesisdesign, line 57: are you sure that there will be always at least one item after the split? Otherwise this line will break.

nepython: Yes, the `name` string is supposed to be a **non-empty** string. As long as it's true, this line won't break.
Below is the output of above line with and without docstring.
```python
In [1]: 'test_status_signal_emitted (test_project.tests.test_test_utils.TestUtils)\nTests the catch_signal test utility function'.split('\n')     # with docstring                                                               
Out[1]: 
['test_status_signal_emitted (test_project.tests.test_test_utils.TestUtils)',
 'Tests the catch_signal test utility function']

In [2]: 'test_status_signal_emitted (test_project.tests.test_test_utils.TestUtils)'.split('\n')    # without docstring                                                                                                         
Out[2]: ['test_status_signal_emitted (test_project.tests.test_test_utils.TestUtils)']
```

tests/test_project/tests/test_test_utils.py
```diff
@@ -40,16 +42,18 @@ def test_deep_merge_dicts(self):
         self.assertDictEqual(deep_merge_dicts(dict1, dict2), merged)
 
     def test_time_logging_runner(self):
+        setattr(settings, 'OPENWISP_SLOW_TEST_THRESHOLD', [0.0, 0.0])
```
nemesisdesign, line 45: This is fragile: if an assertion fails, the last line won't be executed and it will change the behavior of this feature for all the tests, you should always use `mock.patch` or `override_settings` (from django) in these cases.

nepython: `override_settings` works great. Thanks! :smile:

PR #128, opened 2020-08-09
[minor/fix] Running commitcheck issue #127

### LINKED ISSUE #127: [qa] Commit check doesn't seem to work anymore on pull requests
The commit used in https://github.com/openwisp/openwisp-utils/pull/126 should cause the check to fail, but it seems the check is not being run.

atb00ker:
Closes #127 

P.S: This PR made me realize some cleanup is required in this repository, will do that tomorrow! :smile:

### DIFF
--- a/openwisp-qa-check
+++ b/openwisp-qa-check
@@ -4,7 +4,7 @@ show_help() {
   printf "Usage:\n"
   printf "  openwisp-qa-check --help\n"
   printf "  openwisp-qa-check {--migration-path <path> [--migrations-to-ignore <int>] | --skip-checkmigrations}\n"
-  printf "            {[--message <commit-message>] | --skip-checkcommit | --force-checkcommit}\n"
+  printf "            {[--message <commit-message>] | --skip-checkcommit }\n"
   printf "            {[--migration-module <module>] | --skip-checkmakemigrations}\n"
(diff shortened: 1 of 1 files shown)


atb00ker:
Failing properly, failed example: https://travis-ci.org/github/openwisp/openwisp-utils/jobs/716350788

openwisp-qa-check
```diff
@@ -96,15 +95,15 @@ runblack() {
 # 1. --force-checkcommit is specified
 # 2. This function is invoked from Travis building a Pull Request
 runcheckcommit() {
-  [ -n "$TRAVIS_PULL_REQUEST" ] || TRAVIS_PULL_REQUEST=false
```
atb00ker, line 99: I didn't investigate why this was not running, instead I made it run on local machine & travis both! :smile:

atb00ker:
> Can at least make sure that when we are on travis it doesn't run always but only runs in pull requests?

Done! :smile:

atb00ker:
> I have one doubt though, shouldn't travis now run this check, also in the build of this PR?

Yes it should, upon investigation, I realized travis is still using bash 4; so I have used bash 4 syntax that works for both bash 4 and bash 5.

PR #129, opened 2020-08-14
[deps] Update dependencies + minor changes

atb00ker:


requirements-test.txt
```diff
@@ -1,2 +1,3 @@
 # For testing Dependency loaders
-django_netjsonconfig
+# TODO: Github link used until openwisp-controller 0.8+ is released.
+-e git://github.com/openwisp/openwisp-controller.git#egg=openwisp_controller
```
atb00ker, line 3: I just realized this can't be in the 0.3.0 release, need to wait for controller 0.8 release.

atb00ker (commented):
This PR includes changes for:
+ White space changes for improving readability.
+ use openwisp-controller instead of django_netjsonconfig
+ update dependencies
+ supress openwisp-qa-check unnecessary script outputs. 
+ use of `self.subTests` to help find issues with tests.
+ minor non-functional changes

nemesisdesign (changes requested):
openwisp-controller 0.8 is released so this can now be merged, but there's a conflict to resolve @atb00ker

atb00ker:
Updated! :smile:

setup.py
```diff
@@ -44,16 +44,16 @@
...
+            'black<=20.8b1',
             'flake8<=3.9',
             'isort~=5.0',
-            'readme-renderer~=26.0',
+            'readme-renderer~=28.0',
             'coveralls',  # depends on coverage as well
         ],
         'rest': [
-            'djangorestframework>=3.11,<3.12',
+            'djangorestframework~=3.12',
```
nemesisdesign, line 54: are you sure this doesn't impact negatively the other modules?
I think we should test it on other modules before updating it here. If you want you can help me to open issues like this one in each module which uses DRF: https://github.com/openwisp/openwisp-radius/issues/194

See also https://github.com/openwisp/openwisp-utils/issues/164.

setup.py
```diff
@@ -44,16 +44,16 @@
...
             'django-filter>=2.2.0<2.4.0',
-            'drf-yasg>=1.17,<1.18',
+            'drf-yasg~=1.20',
```
nemesisdesign, line 56: are you sure there's no backward incompatible changes affecting us?

setup.py
```diff
@@ -44,16 +44,16 @@
     install_requires=['django-model-utils>=4.0.0,<4.1.0'],
     extras_require={
         'qa': [
-            'black<=19.10b0',
+            'black<=20.8b1',
```
nemesisdesign, line 47: See https://github.com/openwisp/openwisp-utils/issues/165.
Please set it back to the previous value.

setup.py
```diff
@@ -44,16 +44,16 @@
...
             'flake8<=3.9',
             'isort~=5.0',
-            'readme-renderer~=26.0',
+            'readme-renderer~=28.0',
```
nemesisdesign, line 50: @devkapilbansal please can you double check this change doesn't affect the rstcheck negatively?

.travis.yml
```diff
@@ -4,6 +4,8 @@ cache: pip
 python:
   - "3.6"
   - "3.7"
+  - "3.8"
+  - "3.9"
```
nemesisdesign, line 8: we already ran out of credits so I wouldn't add more for now.

nemesisdesign (changes requested):
I don't remember seeing so many changes in the previous review, there's a few changes that worry me, see below for more details.

setup.py
atb00ker: Okay, got i! 👍🏻

setup.py
atb00ker: No mentioned breaking changes in changelog, only python 3.9 support and a bug fixed is mentioned, check here: https://github.com/pypa/readme_renderer/blob/master/CHANGES.rst#280-2020-10-20

setup.py
atb00ker: I'll upgrade and test it on radius, and ping back here. :smile:

setup.py
atb00ker: This only updates requirements for itself no breaking changes mentioned in the changelog, check here: https://drf-yasg.readthedocs.io/en/stable/changelog.html

atb00ker (commented):
I upgraded all them because the last release was just last week, so I assume the next release would have atleast some gap ensuring enough time to upgrade all the relevant modules.

However, I undo changes as mentioned below:

.travis.yml
atb00ker: tl;dr
Just an update on this credit problem, we can ask for more credits and I've applied for more monthly credits.

Details: 

As it turns out, travis was having performance problems because people were abusing their open-source service, so they made the open-source credits as opt-in.

We ran out of credits on travis-ci because we would now have to apply as open-source projects, they will verify we are indeed an open-source project and allow us more credits. :smile:

I've applied for 10 times more credit per month, let's see how much is allocated! :smile:

setup.py
atb00ker: Done!

setup.py
```diff
@@ -47,13 +47,13 @@
             'black<=19.10b0',
```
atb00ker, line 47: Need to update this as well: https://github.com/openwisp/openwisp-utils/blob/6bdf528dea394562bff23fa0aeab7b6f8627c52a/setup.py#L44

setup.py
devkapilbansal: There are no changes in the part of code that rstcheck uses

setup.py
```diff
@@ -47,13 +47,13 @@
...
             'flake8<=3.9',
             'isort~=5.0',
-            'readme-renderer~=26.0',
+            'readme-renderer~=28.0',
```
pandafy, line 50:
```suggestion
            'readme-renderer~=28.0.0',
```
`readme-renderer~=28.0.` translates to `readme-renderer==28.*`. This has created problems in openwisp-notifications earlier. Please check [compatible release in PEP 440](https://www.python.org/dev/peps/pep-0440/#compatible-release) for more info.

pandafy, line 50: ``suggestion
            'readme-renderer~=28.0.0',
```
`readme-renderer~=28.0.` translates to `readme-renderer==28.*`. This has created problems in openwisp-notifications earlier. Please check [compatible release in PEP 440](https://www.python.org/dev/peps/pep-0440/#compatible-release) for more info.

atb00ker: The problem is not everyone knows about / follows [semver.org](https://semver.org/). This particular project does not follow having a minor version, they only have minor fixes and major releases, [check their changelog](https://github.com/pypa/readme_renderer/blob/master/CHANGES.rst).
Hence, `readme-renderer==28.*` would cover all the minor bug fixes in the `28` version, with `readme-renderer~=28.0.0` we would lose that, what do you think? :smile:

pandafy: I had a bad experience with this which broke dependencies of openwisp-notifications thrice. :sweat_smile: This was my only concern for being cautious. Thanks for sharing changelog of the module. I think we can keep `readme-renderer~=28.0`.

nemesisdesign: Looks like they don't follow semver so what @atb00ker did should be correct. Do you agree @pandafy or do you have additional concerns?

pandafy: What @atb00ker has done is correct. :smile:

nemesisdesign (commented):
@atb00ker is this ready for review?

atb00ker:
Yes! :smile:

nemesisdesign (approved):

PR #141, opened 2020-09-01
[qa] Post build failure summary #109

### LINKED ISSUE #109: [qa] Post build failure summary information as a comment on github
Whenever there's a build failure in a PR, it would be great to know what has failed and post a summary as a comment in the github PR.
I found a blog post which explains how to do it: https://damien.pobel.fr/post/github-api-from-travisci/
But any other updated online resources which work will be fine.

The comment should only have the relevant information about failures an nothing else.

If tests are failing we could simply say "Some test case have failed".

In case `openwisp-qa-check` has failed, we could collect the reason of the failure and post it as a message.

One of the ways in which we could do this is to print errors to standard error (instead of printing to standard output), collect the standard error output in a separate var which is then used to build the comment which will be posted to github. 

I suggest trying this change in your own fork first (enable travis on your fork and test the failures there).

KapilBansal320:
Use Travis buddy in notifications webhooks to get post build failure summary
from travis.

Closes #109

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ python:
   - "3.7"
 
 notifications:
-  # will post a summary as a comment on github in case of failure.
+  # will post a summary as a comment on github in case of failure
   webhooks: https://www.travisbuddy.com/
   on_success:never
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ python:
   - "3.7"
 
 notifications:
-  #To get post build fail summary in case of failure.
+  # will post a summary as a comment on github in case of failure.
   webhooks: https://www.travisbuddy.com/
   on_success:never
 
@@ -28,7 +28,7 @@ install:
   - pip install -U $DJANGO
 
 script:
-  # check here to allow travisbuddy to report error cause in case of failure.
+  # the check is done here to allow travisbuddy to include a detailed failure summary
   - ./run-qa-checks
   - coverage run --source=openwisp_utils runtests.py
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -22,10 +27,8 @@ install:
   - pip install -e .[qa,rest]
   - pip install -U $DJANGO
 
-before_script:
-  - ./run-qa-checks
-
 script:
+  - ./run-qa-checks
```
nemesisdesign, line 31: why this change? Is it necessary? If not, please rollback, we should not change code that works without a good reason.

.travis.yml
```diff
@@ -5,6 +5,11 @@ python:
   - "3.6"
   - "3.7"
 
+notifications:
+  webhooks: https://www.travisbuddy.com/
+  on_success:never
+
+
```
nemesisdesign, line 12: let's avoid this extra blank line please?

.travis.yml
KapilBansal320: I too think that this might not be necessary. But after going through travis build cycle, I came to know that after_failure tests only for the error in the scripts. If build fails due to any other code logic, then we might not be able to get the reason for error.

Too verify the same, I tests this on my own repository and here are the results.
[
travis_error
](url)

.travis.yml
KapilBansal320: Done

.travis.yml
nemesisdesign: ok, great, thanks for the explanation, now makes sense, but please add a brief comment before ./run-qa-checks, eg:

```
# check here to allow travisbuddy to report what went wrong in case of failure
./run-qa-checks
```

KapilBansal320: @nemesisdesign done. I added some comments to give a brief.

Issue #110, opened 2020-05-19
[qa] Enforce conventional commit style


nemesifier:
Read OpenWISP Commit conventions: https://github.com/openwisp/openwisp2-docs/issues/109

I still think we should keep consistency with what we've done until now, so I would use "feature" instead of "feat.
The prefix can be in square brackets, eg: `[feature]`, `[change]`.
Ensure first letter after `]` is capital letter.
We should use the same keywords of conventional commits, with the exception of "feat" which is better spelt "feature", 3 characters are worth it in my opinion in this case.
We should also add `[change]` to the allowed prefixes.

We could then add a commit message template like this: https://thoughtbot.com/blog/better-commit-messages-with-a-gitmessage-template we can try it here first and then replicate it in the other modules.

Maybe we could ditch our own commit check solution in favor of a standard one, I just found this tool which looks really interesting!
https://commitizen-tools.github.io/commitizen/

**This change will help us to generate change logs more easily later on**.

devkapilbansal:
@nemesisdesign can we create pre-commit hooks too to ensure that basic checks works correctly before making a commit. Occasionally new contributors forget to run `openwisp-qa-format` or following flake guidelines due to which build failed.

nemesisdesign:
@devkapilbansal I won't be able to help with this now, it's useful to do it but not more important than releasing new features and helping the community to use the work we've been doing, for these reasons I'm urging all the contributors to focus on the issues in https://github.com/orgs/openwisp/projects/4 at least until we have released openwisp-controller 0.8, openwisp-monitoring 0.1, openwisp-firmware-upgrader 0.1 and openwisp-ipam 0.2.

Issue #108, opened 2020-05-15
[deprecated] Remove openwisp-utils-qa-checks in 0.7.0 


nemesisdesign:
Remove openwisp-utils-qa-checks in version 0.6.0.

Issue #98, opened 2020-04-11
[tests] Create test utility to redirect output


nemesisdesign:
Recently some contributors have been confused by the fact that some tests print output to standard output and/or standard error, so they thought tests failed when tests are passing.

Let's use two context managers from the standard library to create two decorators that we'll use to decorate the test methods which produce output, the context managers are:

- [contextlib.redirect_stdout](https://docs.python.org/3/library/contextlib.html#contextlib.redirect_stdout)
- [contextlib.redirect_stdout](https://docs.python.org/3/library/contextlib.html#contextlib.redirect_stderr)

It should be possible to decorate test methods like the following:

```python
import sys
from openwisp_utils.tests import redirect_stdout, redirect_stderr, redirect_any_output

@redirect_stdout
def test_something_out(self):
    print('this test prints output')

@redirect_stderr
def test_something_err(self):
    print('this test prints some error', file=sys.stderr)

@redirect_any_output
def test_something_out(self):
    print('this test prints output')
    print('this test prints some error', file=sys.stderr)    
```

Sometimes we also need to check the output, so we need a way to access it, eg:

```python
@redirect_stdout(io.StringIO())
def test_something_out(self, output):
    action_generating_output()  # pseudocode
    self.assertIn('expected output', output.getvalue())

@redirect_any_output(stdout=io.StringIO(), stderr=io.StringIO())
def test_something_out(self, stdout, stderr):
    action_generating_output()  # pseudocode
    self.assertIn('expected stdout', stdout.getvalue()) 
    self.assertIn('expected stderr', stderr.getvalue())  
```

Here's a sample decorator which does not use the context managers above but which may adapted to do so: https://gist.github.com/jasonbartz/8480004

As a last step, use this decorator to silence the output of some tests in this module.

As usual, tests and (brief) docs are needed as well.

devkapilbansal:
I will look into this issue

devkapilbansal:
@nemesisdesign please check if this can work 
https://github.com/devkapilbansal/openwisp-utils/commit/56d60be1a7466507599294a3897bb35beb301d16
I don't think that we need context managers for it

PR #145, opened 2020-09-04
[qa] Added RestructuredText checks #144

### LINKED ISSUE #144: [qa] Add ReStructuredText checks to openwisp-qa-check
When publishing to pypi, sometimes the publishing fails because of an error in the README.rst, this may happen once the commit and git tag have been already sent, so it's not ideal.

We should avoid this, it shouldn't be hard:

- we can add it to the qa dependencies this tool https://github.com/mgedmin/restview/
- we can add to the qa check a function that calls the rest check (with possibility to skip this), as follows:
  - if README.rst exists, run the check on it
  - if `docs/source` exists, run the check recursively on it

This would allow to enable this check on all the repos which make use of openwisp-utils!

KapilBansal320:
Added ReStructuredText checks to openwisp-qa-check

Will check recursively for errors in reStructuredText

Closes #144

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -550,7 +550,7 @@ name that should be passed to ``./manage.py makemigrations``, eg:
 ``checkrst``
 ^^^^^^^^^^^^^
 
-Ensures all ReStructuredText files are according to guidelines of pypi.
+Checks the syntax of all ReStructuredText files to ensure they can be published on pypi or using python-sphinx.
 
 Settings
 --------
--- a/README.rst
+++ b/README.rst
@@ -27,7 +27,6 @@ openwisp-utils
 
 .. image:: https://img.shields.io/badge/code%20style-black-000000.svg
    :target: https://pypi.org/project/black/
-   
    :alt: code style: black
 
 ------------
@@ -463,6 +462,7 @@ Shell script to run the following quality assurance checks:
 * `checkcommit <#checkcommit>`_
 * `checkendline <#checkendline>`_
 * `checkpendingmigrations <#checkpendingmigrations>`_
+* `checkrst <#checkrst>`_
 * ``flake8`` - Python code linter
 * ``isort`` - Sorts python imports alphabetically, and seperated into sections
 * ``black`` - Formats python code using a common standard
@@ -546,6 +546,11 @@ be created.
 It accepts an optional ``--migration-module`` flag indicating the django app
 name that should be passed to ``./manage.py makemigrations``, eg:
 ``./manage.py makemigrations $MIGRATION_MODULE``.
+
+``checkrst``
+^^^^^^^^^^^^^
+
+Ensures all ReStructuredText files are according to guidelines of pypi.
 
 Settings
 --------
--- a/openwisp-qa-check
+++ b/openwisp-qa-check
@@ -79,6 +79,11 @@ runflake8() {
     || { echoerr "ERROR: Flake8 check failed!"; FAILURE=1; }
 }
 
+runrstcheck() {
+  checkrst && echo "SUCCESS: ReStructuredText check successful!" \
+    || { echoerr "ERROR: ReStructuredText check failed!"; FAILURE=1; }
+}
+
 runisort() {
   isort . --check-only --diff \
     && echo "SUCCESS: Isort check successful!" \
@@ -126,6 +131,7 @@ SKIP_CHECKMIGRATIONS=false
 SKIP_MAKEMIGRATIONS=false
 SKIP_FLAKE8=false
 SKIP_ISORT=false
+SKIP_RSTCHECK=false
 SKIP_BLACK=false
 SKIP_CHECKCOMMIT=false
 FORCE_CHECKCOMMIT=false
@@ -159,6 +165,9 @@ while [ "$1" != "" ]; do
     ;;
   --skip-black)
     SKIP_BLACK=true
+    ;;
+  --skip-checkrst)
+    SKIP_RSTCHECK=true
     ;;
   --skip-checkcommit)
     SKIP_CHECKCOMMIT=true
@@ -208,6 +217,10 @@ fi
 
 if ! $SKIP_FLAKE8; then
   runflake8
+fi
+
+if ! $SKIP_RSTCHECK; then
+  runrstcheck
 fi
 
 if ! $SKIP_CHECKCOMMIT; then
--- a/openwisp_utils/qa.py
+++ b/openwisp_utils/qa.py
@@ -2,9 +2,14 @@
 Common Quality Assurance checks for OpenWISP modules
 """
 import argparse
+import io
 import os
 import re
 import sys
+from glob import iglob
+
+import docutils
+import readme_renderer.rst as readme_rst
 
 
 def _parse_migration_check_args():
@@ -179,6 +184,39 @@ def check_commit_message():
         if not args.quiet:
             print(body)
         sys.exit(1)
+
+
+def check_rst_files():
+    string_io = io.StringIO()
+    settings_overrides = {}
+    settings_overrides.update(readme_rst.SETTINGS)
+    settings_overrides['report_level'] = 2
+    settings_overrides['warning_stream'] = string_io
+    files = [_ for _ in iglob('./**/*.rst', recursive=True)]
+    for file in files:
+        data = read_rst_file(file)
+        try:
+            docutils.core.publish_string(data, settings_overrides=settings_overrides)
+            clean_body = readme_rst.clean(''.join(data))
+            if clean_body is None:
+                # Raised sometimes when rendered failed to generate html
+                string_io.write("Failed to generate output for rendering")
+                raise ValueError("Output Failed")
+        except Exception:
+            pass
+        errors = string_io.getvalue().strip()
+        if errors:
+            body = 'There are some Errors in your RST file syntax \n\n'
+            for error in errors.splitlines():
+                body += '- {}\n'.format(error)
+            print(body)
+            sys.exit(1)
+
+
+def read_rst_file(filename):
+    with open(filename, 'r') as f:
+        data = f.read()
+        return data
 
 
 def _find_issue_mentions(message):
--- a/setup.py
+++ b/setup.py
@@ -36,6 +36,7 @@ setup(
         'console_scripts': [
             'checkmigrations = openwisp_utils.qa:check_migration_name',
             'checkcommit = openwisp_utils.qa:check_commit_message',
+            'checkrst = openwisp_utils.qa:check_rst_files',
         ],
     },
     scripts=[
@@ -50,6 +51,7 @@ setup(
             'black<=19.10b0',
             'flake8<=3.9',
             'isort~=5.0',
+            'readme-renderer~=26.0',
             'coveralls',  # depends on coverage as well
         ],
         'rest': [
--- a/tests/test_project/tests/test_qa.py
+++ b/tests/test_project/tests/test_qa.py
@@ -2,10 +2,15 @@ import io
 import os
 import sys
 from os import path
-from unittest.mock import patch
+from unittest.mock import Mock, patch
 
 from django.test import TestCase
-from openwisp_utils.qa import check_commit_message, check_migration_name
+from openwisp_utils.qa import (
+    check_commit_message,
+    check_migration_name,
+    check_rst_files,
+    read_rst_file,
+)
 
 MIGRATIONS_DIR = path.join(
     path.dirname(path.dirname(path.abspath(__file__))), 'migrations'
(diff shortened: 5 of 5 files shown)


nemesisdesign (changes requested):
I tried using rstcheck to fix an issue that I was having with pypi not accepting the contents of the README, but it didn't work.

I then tried https://github.com/mgedmin/restview/ which did fix the issue. Can you try integrating that tool instead? It's more reliable.

openwisp-qa-check
```diff
@@ -79,6 +79,11 @@ runflake8() {
     || { echoerr "ERROR: Flake8 check failed!"; FAILURE=1; }
 }
 
+runrestcheck() {
+  rstcheck -r . && echo "SUCCESS: ReST check successful!" \
+    || { echoerr "ERROR: ReST check failed!"; FAILURE=1; }
```
nemesisdesign, line 84: ReST > ReSructuredText

openwisp-qa-check
```diff
@@ -126,6 +131,7 @@ SKIP_CHECKMIGRATIONS=false
 SKIP_MAKEMIGRATIONS=false
 SKIP_FLAKE8=false
 SKIP_ISORT=false
+SKIP_ReSTCHECK=false
```
nemesisdesign, line 134: `SKIP_ReSTCHECK` > `SKIP_RESTCHECK`, if we switch to the other tool, we should call it `SKIP_RESTVIEW`

nemesisdesign (commented):
README needs to be updated too

KapilBansal320:
@nemesisdesign I tried using restview too but it opens a web browser and renders errors there instead of returning error and I user have to go manually to the link for checking if there is an error. But as **rstcheck** is not working too, I will try to find an alternative.

pandafy:
@nemesisdesign @KapilBansal320  I don't think we need to add another third party library to add checks for ReST. There is a `rst2html.py` script bundled with `docutils` package. 

I tried the following command on https://github.com/openwisp/openwisp-notifications/commit/3920a51c573aed1e35478e201585b520298df131 commit

```
rst2html.py README.rst > /dev/null
```

And got this output
```
README.rst:69: (ERROR/3) Error in "code-block" directive:
maximum 1 argument(s) allowed, 4 supplied.

.. code-block:: shell
    pip install openwisp-notifications

README.rst:429: (WARNING/2) Cannot analyze code. No Pygments lexer found for "jinja2".
README.rst:645: (WARNING/2) Cannot analyze code. No Pygments lexer found for "jinja2".
```

On current master of openwisp-notifications, nothing was printed to `stderr` conveying that there are no errors.

@KapilBansal320 you many write a script around the bundled script which looks for ReST files recursively and perform this check. I saw there is an option for returning non-zero exit codes when errors are encountered. There might be already something in the `docutils` package which does the same. I will suggest you to explore [availaible scripts in docutils](https://sourceforge.net/p/docutils/code/HEAD/tree/trunk/docutils/tools/) and check out docutils documentation. I found out this [manpage on "rst2html" you can start from there](https://manpages.debian.org/testing/docutils-common/rst2html.1.en.html). 

I think `restview` library is aimed to render ReST instead of enforcing checks.

KapilBansal320:
Thanks @pandafy for your helpful suggestion. I will try to use it for RestCheck now

nemesisdesign:
@KapilBansal320 @pandafy as long as the problem is solved, I am fine with any solution, really, it doesn't make a difference to me.

What is the problem? Sometimes some errors prevent me (or others) from releasing to pypi, because pypi has a ReStructuredText in place which prevents the upload of the release if the description sent cause rendering errors.

We have a practical example we can use. Until this commit in openwisp-notification https://github.com/openwisp/openwisp-notifications/commit/3920a51c573aed1e35478e201585b520298df131, the syntax had a few issues, so, the way to test this solution is to test it against that commit and ensure the check fails.

Then move to this other commit and ensure the check passes: https://github.com/openwisp/openwisp-notifications/commit/071f1912ccdf50f3881052b083461bd5564c3e8d

The simplest solution we can use to solve this problem is fine to me.

KapilBansal320:
@nemesisdesign changes done as requested but I wonder why build failed for ReStructuredText check as it is not implemented till now.

README.rst
```diff
@@ -462,6 +462,7 @@ Shell script to run the following quality assurance checks:
 * `checkcommit <#checkcommit>`_
 * `checkendline <#checkendline>`_
 * `checkpendingmigrations <#checkpendingmigrations>`_
+* `checkReST`_
```
nemesisdesign, line 465: let's use `checkrest` (uniform case, easier to remember)

README.rst
```diff
@@ -546,6 +547,11 @@ It accepts an optional ``--migration-module`` flag indicating the django app
 name that should be passed to ``./manage.py makemigrations``, eg:
 ``./manage.py makemigrations $MIGRATION_MODULE``.
 
+``checkReST``
+^^^^^^^^^^^^^
+
+Ensures all restructured text files are accordind to guidelines of pypi.
```
nemesisdesign, line 553: restructured text > ReStructuredText

README.rst
```diff
@@ -546,6 +547,11 @@ It accepts an optional ``--migration-module`` flag indicating the django app
 name that should be passed to ``./manage.py makemigrations``, eg:
 ``./manage.py makemigrations $MIGRATION_MODULE``.
 
+``checkReST``
```
nemesisdesign, line 550: let's call it `checkrst` to avoid people confusing it with the concept of REST API

openwisp-qa-check
```diff
@@ -79,6 +79,11 @@ runflake8() {
     || { echoerr "ERROR: Flake8 check failed!"; FAILURE=1; }
 }
 
+runrestcheck() {
```
nemesisdesign, line 82: `runrstcheck`

openwisp-qa-check
nemesisdesign: Let's change this to `SKIP_RST`

openwisp-qa-check
```diff
@@ -160,6 +166,9 @@ while [ "$1" != "" ]; do
   --skip-black)
     SKIP_BLACK=true
     ;;
+  --skip-rest)
```
nemesisdesign, line 169: skip-rst

openwisp_utils/qa.py
```diff
@@ -181,6 +185,38 @@ def check_commit_message():
...
+            errors.append(e)
+    if len(errors):
+        body = (
+            'There are some Errors in your RST file syntax'
+            'which can create problems in future:\n\n'
+        )
+        for error in errors:
+            body += '- {}\n'.format(error)
+        print(body)
+        sys.exit(1)
```
nemesisdesign, line 211: please write a few automated tests for this code to ensure future contributors can change it without risking to break the main functionality

openwisp_utils/qa.py
```diff
@@ -181,6 +185,38 @@ def check_commit_message():
...
+
+
+def read_ReST_file(filename):
```
nemesisdesign, line 214: read_rst_file

nemesisdesign (changes requested):
Great progress, see my comments below.

Question: have you tested this solution on openwisp-notifications as suggested?

KapilBansal320:
> Great progress, see my comments below.
> 
> Question: have you tested this solution on openwisp-notifications as suggested?

Yes, I tested it on both commits as mentioned above.

nemesisdesign (commented):
> > Great progress, see my comments below.
> > Question: have you tested this solution on openwisp-notifications as suggested?
> 
> Yes, I tested it on both commits as mentioned above.

Great, thanks.

I forgot to reply to the build failure subject, travis tests the code in the current branch, so since this branch has the new check in place it will be executed, so it's good, it means that it is working. Please proceed to fix the issue that is highlighted by the check.

KapilBansal320:
Here are the results of checking the function of the given commits.

rstchecks

openwisp_utils/qa.py
KapilBansal320: Pardon, I am not able to understand which type of test cases.

openwisp_utils/qa.py
```diff
@@ -181,6 +185,38 @@ def check_commit_message():
...
+            docutils.core.publish_string(data, settings_overrides=settings_overrides)
+            clean_body = readme_rst.clean(''.join(data))
+            if clean_body is None:
+                raise ValueError("Output Failed")
+        except Exception as e:
+            errors.append(e)
+    if len(errors):
+        body = (
+            'There are some Errors in your RST file syntax '
+            'which can create problems in future:\n\n'
```
nemesisdesign, line 206: Let's keep just:

`'There are some Errors in your RST file syntax:\n\n'`

openwisp_utils/qa.py
nemesisdesign: Tests are here: https://github.com/openwisp/openwisp-utils/blob/master/tests/test_project/tests/test_qa.py

nemesisdesign (changes requested):
Good progress, see how the test coverage is decreasing: https://coveralls.io/builds/33288307

Tests are here: https://github.com/openwisp/openwisp-utils/blob/master/tests/test_project/tests/test_qa.py

After tests are added, we should be ready to start using this.

KapilBansal320:
@nemesisdesign I have added all the required test cases.

openwisp_utils/qa.py
```diff
@@ -181,6 +186,39 @@ def check_commit_message():
...
+        data = read_rst_file(file)
+        try:
+            docutils.core.publish_string(data, settings_overrides=settings_overrides)
+            clean_body = readme_rst.clean(''.join(data))
+            if clean_body is None:
+                raise ValueError("Output Failed")
+        except ValueError:
+            sys.exit(1)
+        except Exception:
+            pass
```
pandafy, line 206: @KapilBansal320 can you please add comments here to explicitly mention why `ValueError` is raised. Also. if `ValueError` is encountered the script will exit abruptly and will not tell user what caused the event and the user will think the script is broken. Please make it print a relevant on the terminal.

setup.py
```diff
@@ -50,6 +51,8 @@
             'black<=19.10b0',
             'flake8<=3.9',
             'isort~=5.0',
+            'docutils~=0.16',
```
pandafy, line 54: `docutils` is a dependency of `readme-render`.Maybe we can skip adding that if it does not requires to be kept locked at a specific version.

openwisp_utils/qa.py
pandafy, line 206: @KapilBansal320 can you please add comments describing when `ValueError` will be raised? Also please make sure that the script writes something on the `stdout` mentioning what happened. From reading the code I think the script will exit abruptly which will make user think, the script is broken when `ValueError` is encountered,

tests/test_project/tests/test_qa.py
```diff
@@ -245,3 +254,44 @@ def test_qa_call_check_commit_message_bump_version(self):
...
+
+    def test_qa_call_check_rst_file_syntax(self):
+        with open(self._test_rst_file, 'a+') as f:
+            f.write(
+                """\
+Test File
+=======
+
+.. code:: python
+"""
```
pandafy, line 290: Can you write it as 
```
f.write(
    'Test File \n'
    '======= \n'
    '.. code:: python'
)
```

I think it will look cleaner.

tests/test_project/tests/test_qa.py
```diff
@@ -245,3 +254,44 @@ def test_qa_call_check_commit_message_bump_version(self):
...
+            )
+        try:
+            check_rst_files()
+        except (SystemExit):
+            pass
+        else:
+            self.fail('SystemExit not raised')
```
pandafy, line 297: Use this for better readability
```
with self.assertRaises(SystemExit):
    check_rst_files()
```

pandafy (changes requested):
There are some few minor changes required. I will test it more and will suggest improvements if any.

README.rst
```diff
@@ -463,6 +462,7 @@ Shell script to run the following quality assurance checks:
 * `checkcommit <#checkcommit>`_
 * `checkendline <#checkendline>`_
 * `checkpendingmigrations <#checkpendingmigrations>`_
+* `checkrst`_
```
pandafy, line 465: Please provide a hyperlink to description as done with other commands.

nemesisdesign (approved):
Great, it works! :+1: :clap: 
Thanks a lot to @pandafy for his support and useful reviews as well! :clap: 

@KapilBansal320 other useful issues in this module you may want to look into:

- https://github.com/openwisp/openwisp-utils/issues/108
- https://github.com/openwisp/openwisp-utils/issues/110
- https://github.com/openwisp/openwisp-utils/issues/98

PS: I improved the description in the docs: https://github.com/openwisp/openwisp-utils/pull/145/commits/dfae570e2410422e9fbba9e3a5482912233b6504

PR #151, opened 2020-09-20
[fix] Fixed error in ReStructuredText Check #150

### LINKED ISSUE #150: [qa] ReStructuredCheck failing in netjsonconfig, different approach is needed
The check is failing with:
```
There are some Errors in your RST file syntax 

- <string>:1: (WARNING/2) "raw" directive disabled.
```

However, I don't think the path we've taken is good.

The repositories which have sphinx documentation already have checks to ensure that documentation compiles and renders successfully, these checks simply generate the sphinx docs and they work reliably.

Now we're mixing things.
I asked to add this check because we need to check that the README files are compatible with pypi.
If we start allowing sphinx syntax in the README, we'll hit again the same issue that some of the sphinx syntax won't work on pypi.

I think the right solution is to separate the two things.

Sphinx documentation checks must be done as we already do: https://github.com/openwisp/netjsonconfig/blob/master/run-qa-checks#L8-L10
This new check should only focus on the README compatibility with pypi, removing the changes we introduced recently to make it work on openwisp-radius.
I would also rename the check to make this more explicit: we can call it `checkreadme` to ensure things don't get confused.

@KapilBansal320 what do you think?

devkapilbansal:
Fixes #150

### DIFF
--- a/openwisp_utils/admin.py
+++ b/openwisp_utils/admin.py
@@ -10,10 +10,7 @@ class TimeReadonlyAdminMixin(object):
     """
 
     def __init__(self, *args, **kwargs):
-        self.readonly_fields += (
-            'created',
-            'modified',
-        )
+        self.readonly_fields += ('created', 'modified')
         super().__init__(*args, **kwargs)
 
 
@@ -147,7 +144,7 @@ class ReceiveUrlAdmin(ModelAdmin):
         )
         baseurl = self.receive_url_baseurl
         if not baseurl:
-            baseurl = '{0}://{1}'.format(self.request.scheme, self.request.get_host(),)
+            baseurl = '{0}://{1}'.format(self.request.scheme, self.request.get_host())
         if self.receive_url_querystring_arg:
             url = '{0}{1}?{2}={3}'.format(
                 baseurl,
--- a/openwisp_utils/qa.py
+++ b/openwisp_utils/qa.py
@@ -10,7 +10,6 @@ from glob import iglob
 
 import docutils
 import readme_renderer.rst as readme_rst
-from rstcheck import ignore_sphinx
 
 
 def _parse_migration_check_args():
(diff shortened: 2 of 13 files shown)


KapilBansal320:
@nemesisdesign please review it.

Here I removed recursive checking on the subdirectories like docs as said. Also, **ignore_sphinx** is removed here. Although I don't rename **checkrst** to **checkreadme** as I was checking other rst files in root directory too. Like CHANGES.rst

openwisp_utils/admin.py
```diff
@@ -147,7 +144,7 @@ def receive_url(self, obj):
         )
         baseurl = self.receive_url_baseurl
         if not baseurl:
-            baseurl = '{0}://{1}'.format(self.request.scheme, self.request.get_host(),)
```
nemesisdesign, line 150: there's many changes like this one which are not related to the goal of the PR, this makes history hard to follow.
Can you please separate these changes into a separate commit?

nemesisdesign (changes requested):
Thanks @KapilBansal320, looks good, but there's an issue, see my comment below and let me know if you can do that change.

openwisp_utils/admin.py
KapilBansal320: @nemesisdesign I break down my commit in two, one for the changes and second for the formatting. Is it correct?

devkapilbansal: > @nemesisdesign please review it.
> 
> Here I removed recursive checking on the subdirectories like docs as said. Also, **ignore_sphinx** is removed here. Although I don't rename **checkrst** to **checkreadme** as I was checking other rst files in root directory too. Like CHANGES.rst

@nemesisdesign I am changing the function to check only for readme instead because when I ran it on openwisp2-docs, it fails for index.rst file (as toctree is a sphinx directive and docutils doesn't recognize it).

devkapilbansal:
> @nemesisdesign please review it.
> 
> Here I removed recursive checking on the subdirectories like docs as said. Also, **ignore_sphinx** is removed here. Although I don't rename **checkrst** to **checkreadme** as I was checking other rst files in root directory too. Like CHANGES.rst

@nemesisdesign I am changing the function to check only for readme instead because when I ran it on openwisp2-docs, it fails for index.rst file (as toctree is a sphinx directive and docutils doesn't recognize it).

nemesisdesign (approved):
Good work @devkapilbansal, thanks :+1:

PR #154, opened 2020-10-25
[Fix] Fixed shifting of html due to main-menu #153

### LINKED ISSUE #153: [bug] Html get shifted left due to main-menu
I don't know wheater it was done intentionally or not but when the cursor is moved to the right side cursor change into pointer due to the main menu which is not visible as shown in the figure.

Screenshot (145)

This is causing some other problem. 

Steps to produce:
1. Click on the invisible portion shown in above figure, it will take you to a different page.
2. Click back button and you will see that html get shifted left due to main menu like this.
Screenshot (146)

niteshsinha17:
Fixes #153

### DIFF
--- a/openwisp_utils/admin_theme/static/admin/css/openwisp.css
+++ b/openwisp_utils/admin_theme/static/admin/css/openwisp.css
@@ -144,7 +144,7 @@ html { overflow-x: hidden }
     overflow: hidden;
     position: absolute;
     z-index: 9999 !important;
-    right: -200px;
+    right: -270px;
     top: 0;
     background-color: #464646;
     z-index: 10;


pandafy (changes requested):
@niteshsinha17 the current solution is not robust and does not works with all screen sizes. I was still able to reproduce the bug even in your patch. 

image

I will suggest to investigate why this happens in the first place.

niteshsinha17:
> @niteshsinha17 the current solution is not robust and does not works with all screen sizes. I was still able to reproduce the bug even in your patch.
Okay I will check

niteshsinha17:
@pandafy I will look on it.

pandafy (approved):
LGTM :+1:

atb00ker (approved):
LGTM! :+1: 

Thanks @niteshsinha17 :smile:

PR #156, opened 2020-11-07
[test] Test Utility to redirect noisy output #98

### LINKED ISSUE #98: [tests] Create test utility to redirect output

devkapilbansal:
Fixes #98

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -463,6 +463,101 @@ In order to switch to this test runner you have set the following in your `setti
 
     TEST_RUNNER = 'openwisp_utils.tests.TimeLoggingTestRunner'
 
+``openwisp_utils.tests.redirect_stdout``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This decorator can be used to redirect standard output produced by tests to somewhere else.
+
+Usage example as a decorator:
+
+.. code-block:: python
+
+
+    from openwisp_utils.tests import redirect_stdout
+
+    @redirect_stdout()
+    def test_something(self):
+        function_generating_error() #pseudo code
+        print("some output")
+
+    @redirect_stdout()
+    def test_something_again(self, captured_ouput):
+        # pseudo code
+        call_function_which_will_produce_output()
+        # now you can create assertions on captured output
+        self.assertIn('expected stdout', captured_ouput.getvalue())
+        # if there are more than one assertions, clear the captured output first
+        captured_error.truncate(0)
+        captured_error.seek(0)
+        # you can create new assertion now
+        self.assertIn('another output', captured_error.getvalue())
+
+** Captured ouput can be defined in arguments if assertion needs to be done on captured output, otherwise they can be omitted.**
+
+** Sting IO instance is used as default and if needed, custom String IO instance can be provided **
+
+``openwisp_utils.tests.redirect_stderr``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Similar to ``redirect_stdout``, this decorator can be used to redirect standard error produced by tests to somewhere else.
+
+Usage example as a decorator:
+
+.. code-block:: python
+
+    from openwisp_utils.tests import redirect_stderr
+
+    @redirect_stderr()
+    def test_error(self):
+        function_generating_error() #pseudo code
+        print("this test will print some error")
+
+    @redirect_stderr()
+    def test_error_again(self, captured_error):
+        # pseudo code
+        call_function_which_will_produce_error()
+        # now you can create assertions on captured error
+        self.assertIn('expected error', captured_error.getvalue())
+        # if there are more than one assertions, clear the captured error first
+        captured_error.truncate(0)
+        captured_error.seek(0)
+        # you can create new assertion now
+        self.assertIn('another expected error', captured_error.getvalue())
+
+** Captured error can be defined in arguments if assertion needs to be done on captured error, otherwise they can be omitted.**
+
+** Sting IO instance is used as default and if needed, custom String IO instance can be provided **
+
+
+``openwisp_utils.tests.redirect_any_output``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This decorator can be used to redirect any output (standard output and standard error) produced by tests to somewhere else.
+
+Usage example as a decorator:
+
+.. code-block:: python
+
+    from openwisp_utils.tests import redirect_any_output
+
+    @redirect_any_output()
+    def test_something_out(self):
+        function_generating_output_and_error() #pseudo code
+        print("some output")
+        print("expected error")
+
+    @redirect_stdout()
+    def test_out_again(self, captured output, captured_error):
+        # pseudo code
+        call_function_which_will_produce_any_output()
+        # now you can create assertions on captured error
+        self.assertIn('expected stdout', captured output.getvalue())
+        self.assertIn('expected stderr', captured_error.getvalue())
+
+** Captured output and Captured error can be defined in arguments if assertion needs to be done on captured error or captured output, otherwise they can be omitted.**
+
+** Sting IO instance is used as default and if needed, custom String IO instance can be provided **
+
 Quality Assurance Checks
 ------------------------
 
@@ -505,10 +600,10 @@ You can do multiple ``checkmigrations`` by passing the arguments with space-deli
 For example, this multiple ``checkmigrations``::
 
     checkmigrations --migrations-to-ignore 3 \
-		    --migration-path ./openwisp_users/migrations/ || exit 1
+            --migration-path ./openwisp_users/migrations/ || exit 1
 
     checkmigrations --migrations-to-ignore 2 \
-		    --migration-path ./tests/testapp/migrations/ || exit 1
+            --migration-path ./tests/testapp/migrations/ || exit 1
 
 Can be changed with::
 
--- a/openwisp_utils/tests.py
+++ b/openwisp_utils/tests.py
@@ -1,3 +1,5 @@
+import io
+import sys
 from contextlib import contextmanager
 from time import time
 from unittest import TextTestResult, mock
@@ -76,3 +78,52 @@ class TimeLoggingTestResult(TextTestResult):
 class TimeLoggingTestRunner(DiscoverRunner):
     def get_resultclass(self):
         return TimeLoggingTestResult
+
+
+class RedirectOutput(object):
+    def __call__(self, function):
+        def wrapped_function(*args, **kwargs):
+
+            if hasattr(self, 'stdout'):
+                sys.stdout = self.stdout
+
+            if hasattr(self, 'stderr'):
+                sys.stderr = self.stderr
+
+            try:
+                if hasattr(self, 'stdout') and hasattr(self, 'stderr'):
+                    function(*args, self.stdout, self.stderr, **kwargs)
+                elif hasattr(self, 'stderr'):
+                    function(*args, self.stderr, **kwargs)
+                else:
+                    function(*args, self.stdout, **kwargs)
+            except TypeError:
+                function(*args, **kwargs)
+            finally:
+                if hasattr(self, 'stdout'):
+                    self.stdout.close()
+                    sys.stdout = sys.__stdout__
+                if hasattr(self, 'stderr'):
+                    self.stderr.close()
+                    sys.stderr = sys.__stderr__
+
+        return wrapped_function
+
+
+class redirect_stdout(RedirectOutput):
+    def __init__(self, stdout=None, **kwargs):
+        self.kwargs = kwargs
+        self.stdout = io.StringIO() if stdout is None else stdout
+
+
+class redirect_stderr(RedirectOutput):
+    def __init__(self, stderr=None, **kwargs):
+        self.kwargs = kwargs
+        self.stderr = io.StringIO() if stderr is None else stderr
+
+
+class redirect_any_output(RedirectOutput):
+    def __init__(self, stdout=None, stderr=None, **kwargs):
+        self.kwargs = kwargs
+        self.stdout = io.StringIO() if stdout is None else stdout
+        self.stderr = io.StringIO() if stderr is None else stderr
--- a/tests/test_project/tests/test_qa.py
+++ b/tests/test_project/tests/test_qa.py
@@ -1,6 +1,4 @@
-import io
 import os
-import sys
 from os import path
 from unittest.mock import Mock, patch
 
@@ -11,6 +9,7 @@ from openwisp_utils.qa import (
     check_rst_files,
     read_rst_file,
 )
+from openwisp_utils.tests import redirect_stderr, redirect_stdout
 
 MIGRATIONS_DIR = path.join(
     path.dirname(path.dirname(path.abspath(__file__))), 'migrations'
@@ -46,6 +45,7 @@ class TestQa(TestCase):
             except (SystemExit, Exception) as e:
                 self.fail(e)
 
+    @redirect_stderr()
     def test_qa_call_check_migration_name_failure(self):
         options = [
             [
@@ -68,11 +68,10 @@ class TestQa(TestCase):
                 else:
                     self.fail('SystemExit not raised')
 
-    def test_migration_failure_message(self):
+    @redirect_stdout()
+    def test_migration_failure_message(self, captured_output):
         bad_migration = ['checkmigrations', '--migration-path', MIGRATIONS_DIR]
         with patch('argparse._sys.argv', bad_migration):
-            captured_output = io.StringIO()
-            sys.stdout = captured_output  # redirect stdout
             try:
                 check_migration_name()
             except (SystemExit):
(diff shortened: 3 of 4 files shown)


devkapilbansal:
@atb00ker @nemesisdesign please review it again

openwisp_utils/tests.py
```diff
@@ -76,3 +77,79 @@ def stopTestRun(self):
 class TimeLoggingTestRunner(DiscoverRunner):
     def get_resultclass(self):
         return TimeLoggingTestResult
+
+
+class redirect_stdout(object):
```
devkapilbansal, line 82: All these three functions, i.e., `redirect_stdout`, 'redirect_stderr` and `redirect_any_output` are similar to one another. Is there any way we can inherit them from one base function

tests/test_project/tests/test_qa.py
```diff
@@ -1,6 +1,5 @@
 import io
```
atb00ker, line 1: I see 
```
usage: commitcheck [-h] --message MESSAGE [--quiet]
commitcheck: error: the following arguments are required: --message
...usage: checkmigrations [-h] --migration-path MIGRATION_PATH
                       [--migrations-to-ignore MIGRATIONS_TO_IGNORE] [--quiet]
checkmigrations: error: the following arguments are required: --migration-path
```

in the output, can you please supress that as well?

openwisp_utils/tests.py
```diff
@@ -76,3 +77,70 @@ def stopTestRun(self):
...
+                function(*args, self.stderr ** kwargs)
+
+            finally:
+                self.stderr.close()
+                sys.stderr = sys.__stderr__
+
+        return wrapped_function
+
+
+class redirect_any_output(object):
```
atb00ker, line 124: Could you please document these classes in `README.md`? Please include how to use them.

openwisp_utils/tests.py
```diff
@@ -76,3 +77,70 @@ def stopTestRun(self):
 class TimeLoggingTestRunner(DiscoverRunner):
     def get_resultclass(self):
         return TimeLoggingTestResult
+
+
+class redirect_stdout(object):
```
atb00ker, line 82: The name of this class doesn't follow python class naming conventions, however, I have seen some decorators to not follow it either.

@nemesisdesign, do you know if class name conventions are followed for decorator classes or not?

openwisp_utils/tests.py
atb00ker: Yes, we can create a base class with `__call__` 

where we write code like:

```
if (self.stderr):
   sys.stderr = self.stderr
```
and inherit different classes out of with different `__init__()` 

This should work. Should I make an example?

atb00ker (commented):
Good progress, please see the comments below:

atb00ker:
> @atb00ker @nemesisdesign please review it. If it's fine I will wrote the documentation part

You can proceed with documentation with the work. :smile:

openwisp_utils/tests.py
devkapilbansal: Yes, please it would be helpful

tests/test_project/tests/test_qa.py
devkapilbansal: Okay, I will

openwisp_utils/tests.py
atb00ker: Sure, feel free to update them in the meanwhile and please add the docs so that it's easier for me to use them! :smile:

tests/test_project/tests/test_qa.py
devkapilbansal: I try to supress them as well but seems decorator is not working on argument parser :thinking:

atb00ker: I see, could you investigate on this?

tests/test_project/tests/test_qa.py
```diff
@@ -267,10 +260,9 @@ def test_qa_call_check_rst_file_clean_failure(self):
             pass
         else:
             self.fail('SystemExit not raised')
-        finally:
-            sys.stdout = sys.__stdout__  # reset redirect
 
-    def test_qa_call_check_rst_file_syntax(self):
+    @redirect_stdout(io.StringIO())
```
atb00ker, line 264: I think `io.StringIO()` can be the default argument value, so that we don't have to pass it everytime on every call!

README.rst
```diff
@@ -442,6 +442,74 @@ In order to switch to this test runner you have set the following in your `setti
...
+
+    @redirect_stdout(io.StringIO())
+    def some_test(self):
+    # Pseudo Code
+    print("some output")
+
+    @redirect_stdout(io.StringIO())
+    def some_test(self, stdout):
+    # Pseudo Code
+    self.assertIn('expected stdout', stdout.getvalue())
```
atb00ker, line 465: Add spaces before this line to indicate it's inside the function.

README.rst
```diff
@@ -442,6 +442,74 @@ In order to switch to this test runner you have set the following in your `setti
...
+
+.. code-block:: python
+
+    import io
+    from openwisp_utils.tests import redirect_stdout
+
+    @redirect_stdout(io.StringIO())
+    def some_test(self):
+    # Pseudo Code
+    print("some output")
```
atb00ker, line 460: Add spaces before this line to indicate it's inside the function.

README.rst
```diff
@@ -442,6 +442,74 @@ In order to switch to this test runner you have set the following in your `setti
...
+
+.. code-block:: python
+
+    import io
+    from openwisp_utils.tests import redirect_stderr
+
+    @redirect_stderr(io.StringIO())
+    def some_test(self):
+    # Pseudo Code
+    print("this test will print some error")
```
atb00ker, line 482: Add spaces before this line to indicate it's inside the function.

README.rst
```diff
@@ -442,6 +442,74 @@ In order to switch to this test runner you have set the following in your `setti
...
+
+    @redirect_stderr(io.StringIO())
+    def some_test(self, stderr):
+    # Pseudo Code
+    self.assertIn('expected error', stderr.getvalue())
```
atb00ker, line 487: Add spaces before this line to indicate it's inside the function.

README.rst
```diff
@@ -442,6 +442,74 @@ In order to switch to this test runner you have set the following in your `setti
...
+
+.. code-block:: python
+
+    import io
+    from openwisp_utils.tests import redirect_any_output
+
+    @redirect_any_output(io.StringIO(), io.StringIO())
+    def some_test(self):
+    # Pseudo Code
+    print("some output")
```
atb00ker, line 504: Add spaces before this line to indicate it's inside the function.

README.rst
```diff
@@ -442,6 +442,74 @@ In order to switch to this test runner you have set the following in your `setti
...
+This decorator can be used to redirect any output produced by tests to somewhere else.
+
+Usage example as a decorator:
+
+.. code-block:: python
+
+    import io
+    from openwisp_utils.tests import redirect_any_output
+
+    @redirect_any_output(io.StringIO(), io.StringIO())
```
atb00ker, line 501: maybe try using `redirect_any_output` for solving this problem: https://github.com/openwisp/openwisp-utils/pull/156/files#r519339479

openwisp_utils/tests.py
```diff
@@ -76,3 +77,67 @@ def stopTestRun(self):
 class TimeLoggingTestRunner(DiscoverRunner):
     def get_resultclass(self):
         return TimeLoggingTestResult
+
+
+class RedirectOutput(object):
```
atb00ker, line 82: Thing this can be shortened:

```
def __call__(self, function):
        def wrapped_function(*args, **kwargs):

            if self.stdout:
                sys.stdout = self.stdout
            if self.stderr:
                sys.stderr = self.stderr

                try:
                    if self.stdout and self.stderr:
                         function(*args, self.stdout, self.stderr, **kwargs)
                    elif self.stdout:
                         function(*args, self.stdout, **kwargs)
                    else:
                         function(*args, self.stderr, **kwargs)
                except Exception:
                    function(*args, **kwargs)
                finally:
                    self.stdout.close()
                    if self.stdout:
                         sys.stdout = sys.__stdout__
                    if self.stderr:
                         sys.stderr = sys.__stderr__

        return wrapped_function

```

atb00ker (commented):
Some changes:

tests/test_project/tests/test_qa.py
devkapilbansal: If I try to give it as default, it is giving error as `ValueError: I/O operation on closed file`

README.rst
devkapilbansal: There was an error in `redirect_stderr` which was causing it to print on terminal.

tests/test_project/tests/test_qa.py
devkapilbansal: Should I use io.StringIO() only, if that's the case I can define it in decorator but then we may not be able to provide different output stream other than StringIO

devkapilbansal: Done

openwisp_utils/tests.py
devkapilbansal: Thanks @atb00ker for your guidance.

openwisp_utils/tests.py
```diff
@@ -76,3 +78,55 @@ def stopTestRun(self):
...
+                if hasattr(self, 'stderr'):
+                    self.stderr.close()
+                    sys.stderr = sys.__stderr__
+
+        return wrapped_function
+
+
+class redirect_stdout(RedirectOutput):
+    def __init__(self, stdout=None, **kwargs):
+        """Init and set vars"""
```
nemesisdesign, line 115: this docstring init and set vars is not much useful, same for other occurrences

README.rst
```diff
@@ -442,6 +442,74 @@ In order to switch to this test runner you have set the following in your `setti
...
+
+.. code-block:: python
+
+    import io
+    from openwisp_utils.tests import redirect_stdout
+
+    @redirect_stdout(io.StringIO())
+    def some_test(self):
+        # Pseudo Code
+        print("some output")
```
nemesisdesign, line 460: why are there two tests with the same name? Is it intentional or not? I would use different names.

Applies to other occurrences as well.

README.rst
```diff
@@ -442,6 +442,74 @@ In order to switch to this test runner you have set the following in your `setti
...
+This decorator can be used to redirect standard output produced by tests to somewhere else.
+
+Usage example as a decorator:
+
+.. code-block:: python
+
+    import io
+    from openwisp_utils.tests import redirect_stdout
+
+    @redirect_stdout(io.StringIO())
```
nemesisdesign, line 457: I understand `io.StringIO()` is not necessary (I see you call the decorator in the tests without doing this), therefore I suggest showing examples in a way that users won't be lead to believe they have to do this.

I would then mention that StringIO is used behind the scenes to capture output and if users need they can pass their own StringIO instance.

Applies to other occurrences.

README.rst
```diff
@@ -442,6 +442,74 @@ In order to switch to this test runner you have set the following in your `setti
...
+    def some_test(self, stdout):
+        # Pseudo Code
+        self.assertIn('expected stdout', stdout.getvalue())
+
+``openwisp_utils.tests.redirect_stderr``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Similar to redirect_stdout, this decorator can be used to redirect standard error produced by tests to somewhere else.
```
nemesisdesign, line 470: please wrap ``redirect_stdout`` to mark it as code sample:

```
``redirect_stdout``
```

README.rst
```diff
@@ -442,6 +442,74 @@ In order to switch to this test runner you have set the following in your `setti
...
+    from openwisp_utils.tests import redirect_stdout
+
+    @redirect_stdout(io.StringIO())
+    def some_test(self):
+        # Pseudo Code
+        print("some output")
+
+    @redirect_stdout(io.StringIO())
+    def some_test(self, stdout):
+        # Pseudo Code
```
nemesisdesign, line 464: Instead of pseudo code, I would write something like:

```python
# pseudo code
call_function_which_will_produce_output()
# now you can create assertions on captured output
self.assertIn('expected stdout', stdout.getvalue())
```

Applies to other occurrences as well.

README.rst
```diff
@@ -442,6 +442,74 @@ In order to switch to this test runner you have set the following in your `setti
...
+
+Usage example as a decorator:
+
+.. code-block:: python
+
+    import io
+    from openwisp_utils.tests import redirect_stderr
+
+    @redirect_stderr(io.StringIO())
+    def some_test(self):
```
nemesisdesign, line 480: I see here you're defining the test method without arguments and it seems this works.
I would make this explicit with a comment (the captured output can be defined in the arguments if assertion need to be done on captured output, otherwise they can be omitted), applies to other occurrences as well.

README.rst
```diff
@@ -442,6 +442,74 @@ In order to switch to this test runner you have set the following in your `setti
...
+
+    @redirect_stderr(io.StringIO())
+    def some_test(self, stderr):
+        # Pseudo Code
+        self.assertIn('expected error', stderr.getvalue())
+
+``openwisp_utils.tests.redirect_any_output``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This decorator can be used to redirect any output produced by tests to somewhere else.
```
nemesisdesign, line 492: I recommend make it more explicit: `any output (standard output and standard error)`

nemesisdesign (changes requested):
Great progress! @devkapilbansal :+1: 

I don't see tests for stderr and any_output which assert on the captured output, I think we must ensure assertions are present to ensure the functionality will not be broken inadvertently by future maintenance.

I also left more comments to improve it further.

devkapilbansal:
Hi, @atb00ker @nemesisdesign I use decorator to decorate some more function but getting assertion error

When the function is like the following it works fine
```
    @redirect_stdout()
    def test_print_color(self, captured_output):
        print_color('This is the printed in Red Bold', color_name='red_bold')
        expected = '\033[31;1mThis is the printed in Red Bold\033[0m\n'
        self.assertEqual(captured_output.getvalue(), expected)
        with captured_stdout() as stdout:
            print_color('This is the printed in Red Bold', color_name='red_bold', end='')
        expected = '\033[31;1mThis is the printed in Red Bold\033[0m'
        self.assertEqual(stdout.getvalue(), expected)
        with captured_stdout() as stdout:
            print_color('This is the printed in Red Bold', color_name='invalid')
        expected = '\033[0mThis is the printed in Red Bold\033[0m\n'
        self.assertEqual(stdout.getvalue(), expected)
```

But when I change it the test fails.
```
    @redirect_stdout()
    def test_print_color(self, captured_output):
        print_color('This is the printed in Red Bold', color_name='red_bold')
        expected = '\033[31;1mThis is the printed in Red Bold\033[0m\n'
        self.assertEqual(captured_output.getvalue(), expected)
        print_color('This is the printed in Red Bold', color_name='red_bold', end='')
        expected = '\033[31;1mThis is the printed in Red Bold\033[0m'
        self.assertEqual(captured_output.getvalue(), expected)
        print_color('This is the printed in Red Bold', color_name='invalid')
        expected = '\033[0mThis is the printed in Red Bold\033[0m\n'
        self.assertEqual(captured_output.getvalue(), expected)
```

devkapilbansal:
> Great progress! @devkapilbansal 
> 
> I don't see tests for stderr and any_output which assert on the captured output, I think we must ensure assertions are present to ensure the functionality will not be broken inadvertently by future maintenance.
> 
> I also left more comments to improve it further.

I include tests for any_output but don't see a place where stderr can be used. Should I include a test function myself? :thinking:

nemesisdesign (approved):
Great work @devkapilbansal, I'll be adding some improvements, will send you more info.

PR #160, opened 2020-11-18
[change] Skip commit check on commits prefixed with "Bump" #159

### LINKED ISSUE #159: [change] Skip commit check on commits prefixed with "Bump"
Pull requests created by dependabot fail the commit check: https://github.com/openwisp/openwisp-wifi-login-pages/pull/108
This makes reviewing these PRs timeconsuming because they all fail, also if the tests pass.

In the case of openwisp-wifi-login-pages, we have headless tests running, so if the build pass that should be a good indicator that we can merge.

Purhan:
Closes #159

### DIFF
--- a/openwisp_utils/qa.py
+++ b/openwisp_utils/qa.py
@@ -96,6 +96,7 @@ def check_commit_message():
         # releases
         r'^[A-Za-z0-9.]* release$',
         r"^Bumped VERSION to (.*?)$",
+        r"^Bump (.*?) from (.*?) to (.*?)$",
     ]
     for case in skip_cases:
         if re.match(case, short_desc):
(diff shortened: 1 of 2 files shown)


openwisp-qa-check
```diff
@@ -101,7 +101,7 @@ runblack() {
 
 runcheckcommit() {
   if [ -z "$COMMIT_MESSAGE" ]; then COMMIT_MESSAGE=$(git log -1 --pretty=%B); fi
-  if [ "$TRAVIS" = true ] && [ "$TRAVIS_PULL_REQUEST" = false ]; then
+  if [[ "$COMMIT_MESSAGE" =~ ^Bump ]] || ([ "$TRAVIS" = true ] && [ "$TRAVIS_PULL_REQUEST" = false ]); then
```
nemesisdesign, line 104: look here: https://github.com/openwisp/openwisp-utils/blob/master/openwisp_utils/qa.py#L98
That's where we should do it, please also add a test for it.

nemesisdesign (changes requested):
Thanks for taking this @Purhan, we should use a different approach though, see my comment below.

openwisp-qa-check
Purhan: Thanks for reviewing, @nemesisdesign please take a look now.

nemesisdesign (approved):
Perfect, thanks!

PR #168, opened 2021-01-12
[admin] Added custom dashboard

pandafy:
Added custom landing page in Django admin which can render
different types charts based on the configurable schema.

-----------------------------------------------------

Screenshot from 2021-01-12 23-33-35

The code was able to generate above charts after I populated the database as follows:

1. Create 3 different projects
2. Add operators to these projects.


openwisp_utils/admin_theme/admin.py
```diff
@@ -18,6 +23,72 @@ class OpenwispAdminSite(admin.AdminSite):
     )
     enable_nav_sidebar = False
 
+    def login(self, *args, **kwargs):
+        from django.http import HttpResponseRedirect
+        from django.urls import reverse
```
nemesisdesign, line 28: why these imports here and not at the top?

openwisp_utils/admin_theme/admin.py
```diff
@@ -18,6 +23,72 @@ class OpenwispAdminSite(admin.AdminSite):
...
+        if isinstance(response, HttpResponseRedirect):
+            response = HttpResponseRedirect(reverse('admin:ow_dashboard'))
+        return response
+
+    def ow_dashboard(self, request, extra_context=None):
+        context = {
+            'is_popup': False,
+            'has_permission': True,
+        }
+        from .schema import DASHBOARD_SCHEMA
```
nemesisdesign, line 40: same

openwisp_utils/admin_theme/admin.py
pandafy: I have moved them to the top. In Django's source code, some imports were done in a similar way like for [`get_urls`](https://github.com/django/django/blob/e0a46367df8b17905f1c78f5c86f88d21c0f2b4d/django/contrib/admin/sites.py#L241)

openwisp_utils/admin_theme/admin.py
```diff
@@ -18,6 +27,70 @@ class OpenwispAdminSite(admin.AdminSite):
...
+                raise ImproperlyConfigured(
+                    f'Error adding dashboard element {key}.'
+                    f'REASON: {app_label}.{model_name} could not be loaded.'
+                )
+
+            # Filter query according to organization of user
+            if hasattr(model, 'organization_id') and not request.user.is_superuser:
+                qs = model.objects.filter(
+                    organization_id__in=request.user.organizations_managed
+                )
```
nemesisdesign, line 57: In the case of the monitoring module, the query will be something like:

```python
DeviceMonitoring.objects.filter(device__organization_pk__in=request.user.organizations_managed)
```

So this part needs to be improved to allow modules to supply the information on where to look for the organization field.

One possibility is to add another argument to `query_params`:

- `organization_lookup`, which defaults to `organization__in` but can be overridden, eg:

```python
register_dashboard_element(
    'Device Health Status',
    {
        'query_params': {
            'app_label': 'device_monitoring',
            'model': 'DeviceMonitoring',
            'group_by': 'status',
            'organization_lookup': 'device__organization__in',
        },
        'colors': {
            'unknown': 'gray'
            'ok': 'green',
            'problem': 'yellow',
            'critical': 'red',
        },
    },
)
```

The code can then become something like:

```python
organization_lookup = value['query_params'].pop('organization_lookup', 'organization__in')
app_label, model_name, group_by = value['query_params'].values()
qs = model.objects.filter(**{organization_lookup: request.user.organizations_managed})
```

openwisp_utils/admin_theme/admin.py
```diff
@@ -18,6 +27,70 @@ class OpenwispAdminSite(admin.AdminSite):
...
+            try:
+                model = load_model(app_label, model_name)
+            except ImproperlyConfigured:
+                raise ImproperlyConfigured(
+                    f'Error adding dashboard element {key}.'
+                    f'REASON: {app_label}.{model_name} could not be loaded.'
+                )
+
+            # Filter query according to organization of user
+            if hasattr(model, 'organization_id') and not request.user.is_superuser:
```
nemesisdesign, line 54: I would avoid the if since at the moment we only plan to show elements which are bound to organizations so we should focus on that

nemesisdesign (changes requested):
Great progress @pandafy.

I tested it and looked at the code, we're on the right track, see my comments below regarding areas that have to be improved, especially the `organization_lookup` thing.

Regarding aesthetics, can you do the following?

- show the legend below the pie chart
- ensure the title of the chart (eg: `Operator Project Distribution`) has some white space above so it's not so close to the pie chart, or alternatively, it can be shown on top, which may also make the chart more readable (provided there would be some whitespace below)
- I see the chart is clickable, but there's no visual indication of this, is it possible to use CSS to make the chart show the pointer cursor? I tried setting `cursor:pointer` to `path.surface` using the browser console and it worked, I wonder if it's possible to do the same with CSS

openwisp_utils/admin_theme/admin.py
pandafy: @nemesisdesign  in openwisp_monitoring I was thinking to write define element configuration as below:

```python
register_dashboard_element(
    'Monitoring Status',
    {
        'query_params': {
            'app_label': 'config',
            'model': 'device',
            'group_by': 'monitoring__status',
        },
        'colors': {
            'unknown': 'grey',
            'ok': 'green',
            'critical': 'orange',
            'problem': 'red',
        },
    },
) 
```

Similarly for openwisp-controller it would have become:

```python
register_dashboard_element(
    'Configuration Status',
    {
        'query_params': {
            'app_label': 'config',
            'model': 'device',
            'group_by': 'config__status',
        },
        'colors': {'applied': 'green', 'modified': 'orange', 'error': 'red'},
    },
)
```
-----------

I know it will create a join in the database which will increase the time of execution of the query, but I wanted to keep it simple. The admin class of `Device` in both openwisp-controller and openwisp-monitoring has already implemented filters on based of `config__status` and `monitoring__status` respectively. If we use `Config` or `DeviceMonitoring` models in the element configuration, we will have to pass other arguments as well like where to redirect while clicking on a section of pie chart. 

The current implementation will work with above code. 

------------------

I have opened PRs in openwisp-controller and openwisp-monitoring for adding elements to the dashabord:

- https://github.com/openwisp/openwisp-controller/pull/369
- https://github.com/openwisp/openwisp-monitoring/pull/264

openwisp_utils/admin_theme/admin.py
pandafy: I think it is better to keep it for future scope. Moreover, I don't think the graph added by the `test_project` will be able to work without this condition.

openwisp_utils/admin_theme/admin.py
```diff
@@ -18,6 +27,70 @@ class OpenwispAdminSite(admin.AdminSite):
...
+                values.append(obj['count'])
+            if 'colors' in value:
+                for label in labels:
+                    colors.append(value['colors'][label])
+            value['query_params'] = {'values': values, 'labels': labels}
+            value['colors'] = colors
+            value['target_link'] = f'/admin/{app_label}/{model_name}/?{group_by}='
+
+        context.update({'dashboard_schema': dict(schema)})
+        return render(request, template_name='admin/ow_dashboard.html', context=context)
```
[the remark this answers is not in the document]
pandafy: `dashboad.css` conflicts with [`dashboard.css` of Django's admin site](https://github.com/django/django/blob/master/django/contrib/admin/static/admin/css/dashboard.css). I used `ow-dashboard` for all files for consistency.

pandafy:
I have made the changes regarding layout of the charts.

openwisp_utils/admin_theme/admin.py
nemesisdesign: I left comments in each PR.

nemesisdesign (commented):
Looks a lot better! :+1: 

The slant text does not convince me though:

Screenshot from 2021-01-19 22-31-37

Is there any alternative?
Looking at the plotly examples, I think having a thicker donut body could work, or alternatively, a basic pie chart:
https://plotly.com/javascript/pie-charts/

nemesisdesign (changes requested):
One request: can we specify the priority/order when registering an item?

That way we can say:

- monitoring goes first
- controller 2nd
- topology 3rd, etc.

Otherwise the order used is the one used in `INSTALLED_APPS` which cannot be edited freely.

pandafy:
@nemesisdesign  in https://github.com/openwisp/openwisp-utils/pull/168/commits/68baa1f70720c01a7175e850ba75dd39957c40e2 I have introduced the possibility of ordering the charts by user. But, I am afraid that I might have over-engineered the implementation. Will you please let me know, whether I should write documentation for the current implementation or implement a simpler solution?

openwisp_utils/admin_theme/schema.py
```diff
@@ -0,0 +1,48 @@
...
+        raise ImproperlyConfigured(
+            'Dashboard element parameters should be type `dict`.'
+        )
+    if element_name in DASHBOARD_SCHEMA:
+        raise ImproperlyConfigured(
+            f'{element_name} is an already registered Dashboard Element.'
+        )
+
+    validated_element_params = _validate_element_params(element_params)
+    DASHBOARD_SCHEMA.update({element_name: validated_element_params})
```
nemesisdesign, line 39: An alternative could be to use the order as the dict key (since the order is more important than the name in this case).
This would allow to avoid having to sort it at each registration, we can just leave it unsorted and then sort it in `OpenwispAdminSite` before it will be used.

```python
register_dashboard_element(
    position=1
    params={
        'name': 'Operator Project Distribution',
        'query_params': {
            'app_label': 'test_project',
            'model': 'operator',
            'group_by': 'project__name',
        },
        'colors': {'Utils': 'red', 'User': 'orange'},
    },
)
```

```python
def dashboard(self, request, extra_context=None):
    # sort a copy of DASHBOARD_SCHEMA here based on dict keys
    ...
```

openwisp_utils/admin_theme/admin.py
```diff
@@ -18,6 +28,74 @@ class OpenwispAdminSite(admin.AdminSite):
...
+        ):
+            response = HttpResponseRedirect(reverse('admin:ow_dashboard'))
+        return response
+
+    def dashboard(self, request, extra_context=None):
+        context = {
+            'is_popup': False,
+            'has_permission': True,
+        }
+        schema = copy.deepcopy(DASHBOARD_SCHEMA)
```
nemesisdesign, line 45: I'm not sure if the name `DASHBOARD_SCHEMA`, because more than a schema, it is the configuration of the dashboard. Schema could be confused with the concept of the DB schema or JSON schema, what do you think?

pandafy: I agree with you, we can name it `DASHBOARD_CONFIG`.

openwisp_utils/admin_theme/schema.py
pandafy: > This would allow to avoid having to sort it at each registration, we can just leave it unsorted and then sort it in OpenwispAdminSite before it will be used 

I thought about this but realised that it will introduce a delay while loading the page. All the elements are supposed to be registered through `ready` method of the application. Sorting while adding a new element will only add delay on the startup of the project. After that, there will be no operation needed for sorting the `DASHBOARD_CONFIGURATION`. 

> ```python
> register_dashboard_element(
>     position=1
>     params={
>         'name': 'Operator Project Distribution',
>         'query_params': {
>             'app_label': 'test_project',
>             'model': 'operator',
>             'group_by': 'project__name',
>         },
>         'colors': {'Utils': 'red', 'User': 'orange'},
>     },
> )
> ```

I like using position as key in the dictionary, just like a [TreeMap](https://www.geeksforgeeks.org/treemap-in-java/)

nemesisdesign: I would not worry about this delay @pandafy, since it's negligible, the real delay is caused by the DB queries, that's the part which will most probably hit us with pain in the near future and we'll need to cache, but we'll get back to this issue later on.

If you want to sort each time anyway during registration it's fine as well for me.

README.rst
```diff
@@ -691,6 +806,15 @@ Heading text used in the main ``<h1>`` HTML tag (the logo) of the admin site.
 
 Title shown to users in the index page of the admin site.
 
+``OPENWISP_ADMIN_DASHBOARD_VISIBLE``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
nemesisdesign, line 810: what about renaming this to `_ENABLED`?

pandafy: Done!

pandafy:
@nemesisdesign I will squash these commits once you give green signal. :smile:

nemesisdesign (approved):
Merging in the `dev` branch. Let's continue there.

PR #178, opened 2021-04-03
[qa] Fixed failing commit checks for co-authored commits #167

### LINKED ISSUE #167: [qa] Failing commit checks for co-authored commits
QA checks have started to fail for commits that are co-authored by one or more users.
For example:
1. https://github.com/openwisp/openwisp-controller/runs/1574632899
2. https://github.com/openwisp/ansible-openwisp2/runs/1546631993

### Expected
These checks should not have failed.

devkapilbansal:
Fixes #167

### DIFF
--- a/openwisp_utils/qa.py
+++ b/openwisp_utils/qa.py
@@ -228,7 +228,7 @@ def _find_issue_mentions(message):
      - count of mentions performed correctly
        (using one of the common github keywords)
     """
-    words = message.split(' ')
+    words = message.split()
     issues = []
     issue_locations = []
     counter = 0
@@ -243,7 +243,7 @@ def _find_issue_mentions(message):
         word = words[issue_location - 1]
         issues.append(words[issue_location])
         # check whether issue is just being referred to
-        if issue_location > 2:
+        if issue_location > 1:
             preceding_2words = '{} {}'.format(
                 words[issue_location - 2], words[issue_location - 1]
             )
--- a/tests/test_project/tests/test_qa.py
+++ b/tests/test_project/tests/test_qa.py
@@ -118,6 +118,13 @@ class TestQa(TestCase):
                 '[qa] Improved Y #20\n\n'
                 'Simulation of a special unplanned case\n\n#noqa',
             ],
+            [
+                'commitcheck',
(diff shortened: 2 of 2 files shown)


devkapilbansal:
@nemesisdesign when merging pull request through github interface, a carriage return is added at end of commit message(long description) due to which checks fails.

openwisp_utils/qa.py
```diff
@@ -228,7 +228,7 @@ def _find_issue_mentions(message):
      - count of mentions performed correctly
        (using one of the common github keywords)
     """
-    words = message.split(' ')
+    words = message.split()
```
devkapilbansal, line 231: By removing quotes from here, words will be splitted for any white space (carriage return, new line etc too)

devkapilbansal:
Will fix tests soon

nemesisdesign (changes requested):
can you please add a test for this case?
I had tried it but couldn't replicate it, but I didn't try adding the carriage return in the test text.
See: https://github.com/openwisp/openwisp-utils/blob/42b2500774e9adb23af53b7415f9229a34a86c38/tests/test_project/tests/test_qa.py#L130-L181

tests/test_project/tests/test_qa.py
```diff
@@ -118,6 +118,13 @@ def test_qa_call_check_commit_message_pass(self):
                 '[qa] Improved Y #20\n\n'
                 'Simulation of a special unplanned case\n\n#noqa',
             ],
+            [
+                'commitcheck',
+                '--quiet',
+                '--message',
+                '[fix] Fixed extensibility of openwisp-users and added sample_users test app #377\n\n'
+                'Closes #377\r\n\r\nCo-authored-by: Ajay Tripathi <ajay39in@gmail.com>',
```
devkapilbansal, line 126: This is the commit message I copied from openwisp-controller for which tests are failing

openwisp_utils/qa.py
nemesisdesign: I hadn't noticed this before, thanks!

nemesisdesign (approved):
Thanks @devkapilbansal

PR #182, opened 2021-04-13
[ux] Fix header user on small screens

pniaps:
When width is lower than 1024px, the user goes down to a new row.

https://user-images.githubusercontent.com/7120526/114606419-862ea580-9c9b-11eb-8717-5becc4aac1cc.mp4

This pull request fixes that

https://user-images.githubusercontent.com/7120526/114606449-8dee4a00-9c9b-11eb-82fd-8dfd91a98bb3.mp4

### DIFF
--- a/openwisp_utils/admin_theme/static/admin/css/openwisp.css
+++ b/openwisp_utils/admin_theme/static/admin/css/openwisp.css
@@ -153,7 +153,7 @@ html { overflow-x: hidden }
 #nav-activate:hover { background-color: #000 }
 #nav-activate:focus { outline: none }
 
-#user-tools { padding-right: 50px }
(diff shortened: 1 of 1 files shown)


nemesisdesign (commented):
Thank you @pniaps!

From what I see in the images you have shared, the username is a bit too close to the button, is there a way to maintain the same spacing we have now?

pniaps:
Fixed.
The button was 56x40 but i have changed it to 40x40 (square)
I made 10px separation between user and button, but have i changed it to 24px, exactly as it was.

nemesisdesign (approved):

nemesisdesign:
Thanks!

PR #198, opened 2021-05-21
[fix] Added workaround for minification of browsable API view

pandafy:
django-pipeline strips spaces from pre-formatted text on minifying HTML.
This destroys the representation of data on browsable API views.
Added a workaround to restore presentation to original form using CSS.

Related: https://github.com/openwisp/ansible-openwisp2/issues/256

### DIFF
--- a/openwisp_utils/admin_theme/static/rest_framework/css/browsable-api.css
+++ b/openwisp_utils/admin_theme/static/rest_framework/css/browsable-api.css
@@ -2,6 +2,7 @@
     display: block;
     margin-top: -15px;
     margin-bottom: 10px;
+    white-space: nowrap;
 }
 
 .meta.nocode span {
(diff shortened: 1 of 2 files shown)


openwisp_utils/admin_theme/static/rest_framework/css/browsable-api.css
```diff
@@ -0,0 +1,14 @@
+.meta.nocode {
+    display: block;
+    margin-top: -15px;
+    margin-bottom: 10px;
```
nemesisdesign, line 4: `white-space: nowrap;` is needed here otherwise when there's no minification it will look as follows:

Screenshot from 2021-05-25 20-21-22

I fixed it in the latest commit.

nemesisdesign (approved):

PR #202, opened 2021-05-27
[qa] Added linting of JS and CSS #173

### LINKED ISSUE #173: [QA] Add linting of JS and CSS
Add linting of CSS and JS as we do in [openwisp-controller](https://github.com/openwisp/openwisp-controller/) and [openwisp-notifications](https://github.com/openwisp/openwisp-notifications).

Reference:

- https://github.com/openwisp/openwisp-notifications/blob/d7f20ea41bcff1c0a437b93e98cac2d0a3bbec32/.github/workflows/build.yml#L56-L57
- https://github.com/openwisp/openwisp-notifications/blob/d7f20ea41bcff1c0a437b93e98cac2d0a3bbec32/.stylelintrc.json
- https://github.com/openwisp/openwisp-notifications/blob/d7f20ea41bcff1c0a437b93e98cac2d0a3bbec32/.jshintrc
- https://github.com/openwisp/openwisp-notifications/blob/d7f20ea41bcff1c0a437b93e98cac2d0a3bbec32/run-qa-checks#L4-L5

niteshsinha17:
It also formats ``css`` files when we run ``run-qa-format``
closes #173


openwisp-qa-check
```diff
@@ -60,6 +60,16 @@ runcheckendline() {
   fi
 }
 
+runstylelintcheck() {
+  echo "RUNNING: Stylelint checks"
+  stylelint -f verbose $(find . -type f -name "*.css")
```
atb00ker, line 65: I think this command would spit a lot of unnecessary information even when the test passes.
To avoid it, checkout the how the other functions are suppressing output on success.
Example:
```
runstylelintcheck() {
  stylelint $(find . -type f -name "*.css") &&
    echo "SUCCESS: Stylelint check successful!" ||
    {
      echoerr "ERROR: Stylelint check failed! Hint: did you forget running openwisp-qa-format?"
      FAILURE=1
    }
}
```

openwisp-qa-check
```diff
@@ -60,6 +60,16 @@ runcheckendline() {
...
+}
+
+runjshintcheck() {
+  echo "RUNNING: jshint checks"
+  jshint --verbose $(find . -type f -name "*.js" -a ! -path "*vendor/*.js")
```
atb00ker, line 70: Just like above it will give a lot of output even when the test passes.

```
runjshintcheck() {
  jshint $(find . -type f -name "*.js" -a ! -path "*vendor/*.js") &&
    echo "SUCCESS: Stylelint check successful!" ||
    {
      echoerr "ERROR: Stylelint check failed! Hint: did you forget running openwisp-qa-format?"
      FAILURE=1
    }
}
```

atb00ker (changes requested):
Please test my recommendations in success case and fail case.
This openwisp-qa-check will never fail our CI; hence marking as request changes! :smile:

niteshsinha17:
@atb00ker I have one doubt. I am writing these test at two placed ``openwisp-qa-checks`` and ``run-qa-checks``. If I do not write it in ``run-qa-checks`` it never runs on ``run-qa-checks`` command but other tests run properly. Am I missing something?

atb00ker:
You need to do `pip install .[qa,rest]` to update the script.

openwisp-qa-check
```diff
@@ -188,6 +200,12 @@ while [ "$1" != "" ]; do
   --skip-checkcommit)
     SKIP_CHECKCOMMIT=true
     ;;
+  --skip-stylelint)
+    SKIP_STYLELINT=true
+    ;;
+  --skip-jshint)
+    SKIP_JSHINT=true
```
nemesisdesign, line 207: Great to see that we are adding this in this script! Will be great for reusability.

But we definitely should not do it this way @niteshsinha17 @atb00ker, because it will cause some existing builds to fail because these linters may not be installed.

We must introduce this feature in a backward compatible way, so either by default these linters should not run (the options could be something like ``--css`` and ``--js`` to activate the respective linters) or before running the shell script should verify the availability of the linter program and skip if it's not present.

Moreover, we'll have to update the ``show_help`` function above in this file and the README of the module to mention these options.

nemesisdesign (changes requested):
This commit shouldn't be here: https://github.com/openwisp/openwisp-utils/pull/202/commits/2675cfff298bcf4eac420c9064f013f1896799f8

openwisp-qa-check
niteshsinha17: You are right. I was also thinking about this.

run-qa-checks
```diff
@@ -2,7 +2,10 @@
 set -e
 
 qa-checks () {
-    openwisp-qa-check --migration-path "./tests/test_project/migrations"
+    openwisp-qa-check \
```
niteshsinha17, line 5: @nemesisdesign @atb00ker Its ready for review. We will need to add css and js test for each module like this to run them in CI.

To document about this addition we can mention it here. http://openwisp.io/docs/developer/contributing.html#coding-style-conventions

atb00ker (commented):
Tested, LGTM. :smile:

I just realized.
The optoins `--stylelint` & `--jshint` describe much better what's going to run.
I wouldn't be able to guess what `--css` and `--js` are going to do.

run-qa-checks
nemesisdesign: This section of the README needs to be updated: https://github.com/openwisp/openwisp-utils#openwisp-qa-check

openwisp-qa-check
```diff
@@ -60,6 +68,26 @@ runcheckendline() {
   fi
 }
 
+runstylelintcheck() {
+  stylelint $(find . -type f -name "*.css") &&
```
nemesisdesign, line 72: what happens if `stylelint` is not installed? Can you make sure the function skips with a warning?
We need this to avoid breaking existing builds which are using the master version of this module.

openwisp-qa-check
```diff
@@ -60,6 +68,26 @@ runcheckendline() {
...
+    echo "SUCCESS: Stylelint check successful!" ||
+    {
+      echoerr "ERROR: Stylelint check failed! Hint: did you forget running openwisp-qa-format?"
+      FAILURE=1
+    }
+
```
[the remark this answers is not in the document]
nemesisdesign: is this resolved?

openwisp-qa-check
```diff
@@ -60,6 +68,26 @@ runcheckendline() {
...
+}
+
+runjshintcheck() {
+  jshint --verbose $(find . -type f -name "*.js" -a ! -path "*vendor/*.js") &&
```
nemesisdesign, line 82: what happens if `jshint` is not installed? Can you make sure the function skips with a warning?
We need this to avoid breaking existing builds which are using the master version of this module.

nemesisdesign (changes requested):
Still not ready, see below.

> I just realized.
> The optoins `--stylelint` & `--jshint` describe much better what's going to run.
> I wouldn't be able to guess what `--css` and `--js` are going to do.

@niteshsinha17 what do you think about this suggestion? I'm fine with both nomenclatures.
The advantage of using `css` and `js` is that in the future we change the underlying linting library, we don't need to change the flag name.
Maybe we can use `csslinter` and `jslinter` to make it extra clear?

openwisp-qa-check
niteshsinha17: Added a check for this

openwisp-qa-check
niteshsinha17: Added a check for this

README.rst
```diff
@@ -814,6 +814,7 @@ automated builds of different OpenWISP modules.
 ^^^^^^^^^^^^^^^^^^^^^^
 
 Shell script to automatically format Python code. It runs ``isort`` and ``black``.
+It also runs ``stylelint`` to format CSS code.
```
atb00ker, line 817: Shell script to automatically format Python code. It runs ``isort``. ``black`` and  ``stylelint``.

openwisp-qa-check
```diff
@@ -60,6 +67,45 @@ runcheckendline() {
   fi
 }
 
+runstylelintcheck() {
+  package='stylelint'
+  if which node > /dev/null
+    then
+      if [ `npm list -g | grep -c $package` -eq 0 ]; then
```
atb00ker, line 74: What if I have `npm`? Does `npm` come bundled with `node`?
What do you think about using `whereis stylelint` instead? 
`whereis` comes out of the box and works great in such situations!

atb00ker (approved):
I think it's close to merge! :smile:

README.rst
niteshsinha17: I think its better. 
Shell script to automatically format Python and CSS code. It runs isort. black and stylelint.

openwisp-qa-check
niteshsinha17: I will check this and let you know

niteshsinha17: @atb00ker How should I use ``whereis`` for this situation?

atb00ker: @niteshsinha17 What do you mean? What's the challenge? Can you please elaborate?

niteshsinha17: Yes `` which stylelint`` will work fine. 
I am getting how to write it with ``whereis`` I tried multiple ways but it was not working as expected. Can you please share exact syntax with ``whereis``

atb00ker: Some version of `whereis stylelint | grep /stylelint` should work but don't worry about it, let's focus on `which` only, `whereis` is not a requirement. :smile:

openwisp-qa-format
```diff
@@ -3,3 +3,4 @@ set -e
 
 isort .
 black -S .
+stylelint $(find . -type f -name "*.css") --fix
```
nemesisdesign, line 6: please ensure ``stylelint`` is available before trying to run it

nemesisdesign (changes requested):
The changes to the CSS are working fine, only one detail remaining to handle, see below.

niteshsinha17:
@nemesisdesign and @atb00ker Updated this PR once again.
**Changes made**
1. This time added some more rules in ``stylelint``
2. I am failing checks when ``stylelint`` and ``jshint`` are not installed. 
3. Made changes requested by you last time.

Its ready now

atb00ker (approved):
LGTM! 😄

openwisp-qa-format
```diff
@@ -1,6 +1,12 @@
 #!/bin/bash
 set -e
 
+if which stylelint > /dev/null
+  then
+    stylelint $(find . -type f -name "*.css") --fix
+  else
+    echo "MODULE NOT FOUND: Please install stylelint"
+    exit 1
```
nemesisdesign, line 9: once more time, I think this is wrong. Didn't we agree this feature must be backward compatible?

So why fail if `styleling` is not installed? This command must not fail if stylelint is missing. It could simply say "The program "stylelint" is not installed, Skipping CSS auto-formatting." or something similar.

README.rst
nemesisdesign: Please mention that the command won't fail if ``stylelint`` is not installed but will skip instead.

nemesisdesign (changes requested):

niteshsinha17:
@nemesisdesign I have fixed them. I will be more carefull next time.

nemesisdesign (approved):

openwisp-qa-check
```diff
@@ -60,6 +67,36 @@ runcheckendline() {
   fi
 }
 
+runstylelintcheck() {
+  package='stylelint'
+  if which stylelint > /dev/null; then
+    stylelint verbose $(find . -type f -name "*.css") &&
+      echo "SUCCESS: Stylelint check successful!" ||
+      {
+        echoerr "ERROR: Stylelint check failed! Hint: did you forget running openwisp-qa-format? You may need some manual fixing."
```
niteshsinha17, line 76: I think we can change it. to "ERROR: Stylelint check failed! Hint: did you forget running openwisp-qa-format? You may need to do some manual fixing."

nemesisdesign (approved):

PR #210, opened 2021-06-14
[enhancement] Dashboard extra-config option

atb00ker:
#### Changes

I want to pass custom variables like 'base_url_to_monitoring' in my template, currently I can't add the same in the context.
This PR allows me to add any custom variable to the templates. 

#### Checklist

Tested in https://github.com/openwisp/openwisp-monitoring/pull/304

### DIFF
--- a/openwisp_utils/admin_theme/dashboard.py
+++ b/openwisp_utils/admin_theme/dashboard.py
@@ -214,7 +214,7 @@ def get_dashboard_context(request):
         if 'js' in template_config[0]:
             js += list(template_config[0]['js'])
         if template_config[1]:
-            extra_config.update(template_config[1])
+            extra_config = template_config[1]
 
     context.update(
         {
--- a/openwisp_utils/admin_theme/dashboard.py
+++ b/openwisp_utils/admin_theme/dashboard.py
@@ -66,7 +66,7 @@ def _validate_template_config(config):
     return config
 
 
-def register_dashboard_template(position, config):
+def register_dashboard_template(position, config, extra_config=None):
     """
     Registers a dashboard template
     register_dashboard_template(int, dict)
@@ -79,14 +79,19 @@ def register_dashboard_template(position, config):
         raise ImproperlyConfigured(
             'Dashboard template config parameters should be of type `dict`.'
         )
+    if extra_config and not isinstance(extra_config, dict):
+        raise ImproperlyConfigured(
+            'Dashboard template extra_config parameters should be of type `dict`.'
+        )
+
     if position in DASHBOARD_TEMPLATES:
         raise ImproperlyConfigured(
             f'Cannot register template {config["template"]}. '
             f'Another template is already registered at position n. "{position}": '
-            f'{DASHBOARD_TEMPLATES[position]["template"]}'
+            f'{DASHBOARD_TEMPLATES[position][0]["template"]}'
         )
     validated_config = _validate_template_config(config)
-    DASHBOARD_TEMPLATES.update({position: validated_config})
+    DASHBOARD_TEMPLATES.update({position: [validated_config, extra_config]})
 
 
 def unregister_dashboard_template(path):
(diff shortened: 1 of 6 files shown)


openwisp_utils/admin_theme/dashboard.py
```diff
@@ -198,15 +203,18 @@ def get_dashboard_context(request):
...
-        if 'js' in template_config:
-            js += list(template_config['js'])
+    for _, template_config in DASHBOARD_TEMPLATES.items():
+        templates.append(template_config[0]['template'])
+        if 'css' in template_config[0]:
+            css += list(template_config[0]['css'])
+        if 'js' in template_config[0]:
+            js += list(template_config[0]['js'])
+        if template_config[1]:
+            extra_config.update(template_config[1])
```
nemesisdesign, line 217: I'd just replace it to avoid ambiguity, `extra_config = template_config[1]`, the use of `update()` makes it look as if other parts of the code modify this.

tests/test_project/templates/dashboard_test.html
```diff
@@ -1 +1,2 @@
 <div style="display:none">Testing dashboard</div>
+<div style="display:none">{{ test_extra_config }}</div>
```
nemesisdesign, line 2: Let's add an if so this extra div is added only if needed.

nemesisdesign (changes requested):
Looks good! A couple of minor request for improvement below.

tests/test_project/templates/dashboard_test.html
atb00ker: Hey, @nemesisdesign, this is just the test code for UTs, it's not suppose to be exposed to user and in the UTs, it will always return `true`.
So, do you think adding that would be helpful?

openwisp_utils/admin_theme/dashboard.py
atb00ker: There is a slight problem in this change.
Consider:
In module 1 we have:
```
        register_dashboard_template(
            position=0,
            config={
                'template': 'dashboard_test.html',
                'css': ('dashboard-test.css',),
                'js': ('dashboard-test.js',),
            },
            extra_config={'test_extra_config': 'dashboard-test.config'},
        )
```
in module 2, we have:
```
        register_dashboard_template(
            position=0,
            config={
                'template': 'dashboard_test.html'
            },
            extra_config={'test_extra_config2': 'dashboard-test2.config'},
        )
 ```

Then the value of `test_extra_config` would be lost with `extra_config = template_config[1]`.
I don't think that's desirable. 
`extra_config.update(template_config[1])` fixes that problem! 😄

tests/test_project/templates/dashboard_test.html
nemesisdesign: yes, I think we have to, it's a question of approach: it's not good to include data which is not going to be used, if we allow this kind of practice we'll soon end up with many cases like this with the result that the response data will grow unnecessarily.
We should just stick with best practices whenever possible, unless there's a good reason to do it differently.

openwisp_utils/admin_theme/dashboard.py
nemesisdesign: why would you want to have 2 modules overriding and extending the exra conf? What problem does that solve?

atb00ker: > why would you want to have 2 modules overriding the same template and extending the exra conf? What problem does that solve?

I don't think it's already been done anywhere, but I sure can think that there maybe a template in future which might want behave differently given different arguments and is used multiple times in a page.
In anycase, since it's not required for now, I am changing it to the recommended change.

tests/test_project/templates/dashboard_test.html
atb00ker: Done.

nemesisdesign (approved):
Thanks @atb00ker :+1:

Issue #221, opened 2021-07-30
[ui] Fix alignment of text on logout page.


niteshsinha17:
![image](https://user-images.githubusercontent.com/42765292/127681111-2ce69de1-686f-45bc-aca8-5202e6f5b431.png)

PR #222, opened 2021-07-31
[test] Added selenium tests for main menu #199

### LINKED ISSUE #199: [test] Selenium tests for navigation menu
Write selenium tests to ensure proper working of the navigation menu.
Also, create a checklist to show what is being tested and what is left ( In the PR description )

niteshsinha17:
#### Changes
- Added selenium test for the main menu.
- Fixed some minor issues.
- Fixed this one #221  here because only one line was needed.

#### Checklist

closes #199

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,13 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
 
+    - name: Install Chromium for Selenium testing
+      run: |
+        sudo snap install chromium
+        wget https://chromedriver.storage.googleapis.com/91.0.4472.19/chromedriver_linux64.zip
+        sudo unzip chromedriver_linux64.zip -d /usr/local/bin/
+        sudo chmod +x /usr/local/bin/chromedriver
+
     - name: Install python dependencies
       run: |
         pip install -U pip wheel setuptools
--- a/openwisp_utils/admin_theme/static/admin/css/openwisp.css
+++ b/openwisp_utils/admin_theme/static/admin/css/openwisp.css
@@ -658,6 +658,10 @@ input[type='button']:hover,
   float: none;
   margin-bottom: 26px;
 }
+.no-auth .title-wrapper h1{
+  float: none;
+  margin: 0;
+}
 #changelist .results + .paginator{
   box-sizing: border-box;
   font-size: 14px;
--- a/openwisp_utils/admin_theme/static/admin/js/menu.js
+++ b/openwisp_utils/admin_theme/static/admin/js/menu.js
@@ -125,7 +125,6 @@ function setMenuToggleText(){
 
 function setMenu() {
   setMenuState();
-  setMenuToggleText();
   setTimeout(function () {
     // Transition fix: Add transition to menu and main content
     // after some time.
@@ -161,6 +160,7 @@ function initToggleMenuHandlers() {
   menuBackdrop.addEventListener('click', function (e) {
     e.stopPropagation();
     owContainer.classList.toggle('toggle-menu');
+    setMenuToggleText();
   });
 }
 
--- a/openwisp_utils/admin_theme/templates/admin/base_site.html
+++ b/openwisp_utils/admin_theme/templates/admin/base_site.html
@@ -14,6 +14,16 @@
     <link rel="stylesheet" type="text/css" href="{% static "admin/css/openwisp.css" %}" />
     <link rel="icon" type="image/x-icon" href="{% static "ui/openwisp/images/favicon.png" %}" />
 {% endif %}
+{% if OPENWISP_TEST_MODE %}
+    <link rel="stylesheet" type="text/css" href="{% static "admin/css/openwisp.css" %}" />
+    <link rel="icon" type="image/x-icon" href="{% static "ui/openwisp/images/favicon.png" %}" />
+    {% comment %} disable transitions during testing {% endcomment %}
+    <style>
+      *{
+        transition:none!important;
+      }
+    </style>
+{% endif %}
 {% endblock %}
 
 {% block menu-bar %}
--- a/tests/test_project/tests/test_selenium.py
+++ b/tests/test_project/tests/test_selenium.py
@@ -1,0 +1,319 @@
+from selenium import webdriver
+from selenium.common.exceptions import NoSuchElementException
+from selenium.webdriver import ActionChains
+from selenium.webdriver.common.by import By
+from selenium.webdriver.common.desired_capabilities import DesiredCapabilities
+from selenium.webdriver.support import expected_conditions as EC
+from selenium.webdriver.support.ui import WebDriverWait
+
+from .utils import SeleniumTestCase
+
+
+class TestMenu(SeleniumTestCase):
+    admin_username = 'admin'
+    admin_password = 'password'
+    maxamize_menu = 'Maximize menu'
+    minimize_menu = 'Minimize menu'
+    transition = 'none 0s ease 0s'
+    # none because transition has been set to none during tests
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        chrome_options = webdriver.ChromeOptions()
+        chrome_options.add_argument('--window-size=1366,768')
+        chrome_options.add_argument('--ignore-certificate-errors')
+        chrome_options.add_argument('--remote-debugging-port=9222')
+        capabilities = DesiredCapabilities.CHROME
+        cls.web_driver = webdriver.Chrome(
+            options=chrome_options, desired_capabilities=capabilities
+        )
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.web_driver.quit()
+        super().tearDownClass()
+
+    def tearDown(self):
+        # Clear local storage
+        self.web_driver.execute_script('window.localStorage.clear()')
+
+    def setUp(self):
+        self.admin = self._create_admin(
+            username=self.admin_username, password=self.admin_password
+        )
+
+    def test_addition_of_transition_effect(self):
+        self.login()
+        menu = self.web_driver.find_element_by_id('menu')
+        main_content = self._get_main_content()
+        menu_toggle = self._get_menu_toggle()
+        self.assertEqual(menu.value_of_css_property('transition'), self.transition)
+        self.assertEqual(
+            main_content.value_of_css_property('transition'), self.transition
+        )
+        self.assertEqual(
+            menu_toggle.value_of_css_property('transition'), self.transition
+        )
+
+    def _test_menu_state(self, open, is_narrow=False):
+        logo = self._get_logo()
+        hamburger = self._get_hamburger()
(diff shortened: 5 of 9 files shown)


openwisp_utils/admin_theme/static/admin/css/openwisp.css
```diff
@@ -658,6 +658,10 @@ input[type='button']:hover,
   float: none;
   margin-bottom: 26px;
 }
+.no-auth .title-wrapper h1{
+  float: none;
+  margin: 0;
+}
```
niteshsinha17, line 664: I will be removing these lines. It has been fixed here. https://github.com/openwisp/openwisp-utils/pull/226

nemesisdesign (commented):
Please rebase on the latest master.

tests/test_project/tests/utils.py
```diff
@@ -0,0 +1,160 @@
...
+            driver = self.web_driver
+        if not username:
+            username = self.admin_username
+        if not password:
+            password = self.admin_password
+        driver.get(f'{self.live_server_url}/admin/login/')
+        if 'admin/login' in driver.current_url:
+            driver.find_element_by_name('username').send_keys(username)
+            driver.find_element_by_name('password').send_keys(password)
+            driver.find_element_by_xpath('//input[@type="submit"]').click()
```
atb00ker, line 57: preferably we will have a single source of all strings, i.e if we want to change `/admin/login` it should be a one place change.
I failed to do it in docker-openwisp when I made tests for that, but perhaps we can start having that repository now, so that changes in future are easier, what do you think?

atb00ker (commented):
This is really good, I haven't run these tests on my machine but I like the way you've started them.
Here are some points!

These can be handled independent of GSoC! :smile:

openwisp_utils/admin_theme/static/admin/js/menu.js
```diff
@@ -194,7 +194,6 @@ function setMenuToggleText() {
 
 function setMenu() {
   setMenuState();
-  setMenuToggleText();
   setTimeout(function () {
```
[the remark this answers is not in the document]
niteshsinha17: Because add this moment menu changes its state according to user choice and it looks weird due to transition duration. So adding transition after some time.

tests/test_project/tests/utils.py
niteshsinha17: I agree with you. We can use it in other modules

.github/workflows/ci.yml
```diff
@@ -38,6 +38,15 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
 
+    - name: Install Chromium for Selenium testing
+      run: |
+        sudo apt update
+        sudo apt install snapd
+        sudo snap install chromium
+        wget https://chromedriver.storage.googleapis.com/91.0.4472.19/chromedriver_linux64.zip
```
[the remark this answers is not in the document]
niteshsinha17: Let us do it in separate PR

nemesisdesign (changes requested):
@niteshsinha17 please ensure this is rebased on the latest master.

tests/test_project/tests/test_selenium.py
```diff
@@ -0,0 +1,418 @@
...
+        with self.subTest('Test menu do not remain open on page change or refresh'):
+            self._test_menu_state(False, is_narrow=True)
+        self._test_account_component(is_narrow=True)
+        self._test_menu_dropdown(is_narrow=True)
+        with self.subTest('Test menu on popup page'):
+            self._test_popup_page()
+        self._test_login_and_logout_page()
+        self.web_driver.set_window_size(1366, 768)
+
+    def _test_login_and_logout_page(self, is_logged_in=True):
```
pandafy, line 259: Can we please move these helper methods to the beginning of the class declaration?

pandafy (changes requested):
Good work @niteshsinha17! 👏🏼 
I tested this on my local machine and it works. 

Please see my comments below.

.github/workflows/ci.yml
```diff
@@ -38,6 +38,15 @@ jobs:
...
 
+    - name: Install Chromium for Selenium testing
+      run: |
+        sudo apt update
+        sudo apt install snapd
+        sudo snap install chromium
+        wget https://chromedriver.storage.googleapis.com/93.0.4577.63/chromedriver_linux64.zip
+        sudo unzip chromedriver_linux64.zip -d /usr/local/bin/
+        sudo chmod +x /usr/local/bin/chromedriver
+
```
pandafy, line 49: We don't need to add this. It is [already added in the virtual environment](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#browsers-and-drivers) used by GitHub Actions.

openwisp_utils/admin_theme/templates/admin/base_site.html
```diff
@@ -14,6 +14,16 @@
...
+{% if OPENWISP_TEST_MODE %}
+    <link rel="stylesheet" type="text/css" href="{% static "admin/css/openwisp.css" %}" />
+    <link rel="icon" type="image/x-icon" href="{% static "ui/openwisp/images/favicon.png" %}" />
+    {% comment %} disable transitions during testing {% endcomment %}
+    <style>
+      *{
+        transition:none!important;
+      }
+    </style>
+{% endif %}
```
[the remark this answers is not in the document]
nemesisdesign: definitely! We should be able to do this by extending a specific block and adding the CSS we need.

tests/test_project/tests/utils.py
```diff
@@ -0,0 +1,179 @@
...
+    Get the configurations that are to be used for all the tests.
+    """
+
+    config_file = os.path.join(os.path.dirname(__file__), 'config.json')
+    root_location = os.path.join(os.path.dirname(os.path.realpath(__file__)), '..')
+    with open(config_file) as json_file:
+        config = json.load(json_file)
+
+
+class SeleniumTestCase(StaticLiveServerTestCase, TestConfig):
```
nemesisdesign, line 21: pay attention! It's possible that since this class extends a  django test case class it will be detected as tests to run and will cause migrations to run and slow down the test process without any benefit.

Please double check this and if this is confirmed you could remove the inheritance on `StaticLiveServerTestCase` and rename this `SeleniumTestMixin`. Then you can inherit from ` `StaticLiveServerTestCase` in the actual test classes.

tests/test_project/tests/test_selenium.py
```diff
@@ -0,0 +1,418 @@
...
+            except TimeoutException:
+                is_visible = False
+            self.assertEqual(is_visible, True)
+            mg_head.click()
+            actions.move_to_element(mg_head)
+            actions.perform()
+            self.assertEqual(mg_label.is_displayed(), False)
+
+    def _test_popup_page(self):
+        self.open('/admin/auth/user/add/?_to_field=id&_popup=1')
```
nemesisdesign, line 405: always use `reverse` for URLs

tests/test_project/tests/test_selenium.py
```diff
@@ -0,0 +1,418 @@
...
+        WebDriverWait(self.web_driver, 2).until(
+            EC.visibility_of_element_located(
+                (By.XPATH, self.config['main_content_xpath'])
+            )
+        )
+        with self.assertRaises(NoSuchElementException):
+            self._get_menu()
+        self.open('/admin/')
```
nemesisdesign, line 413: same here and any other occurrence, please use `reverse()` which makes the code resistant to changes.

nemesisdesign (changes requested):
Great work Nitesh, I also left my comments.

tests/test_project/tests/test_selenium.py
niteshsinha17: Adding them just below the setUp func.

tests/test_project/tests/utils.py
niteshsinha17: You are right :+1:

tests/test_project/templates/admin/base_site.html
```diff
@@ -0,0 +1,16 @@
+{% extends 'admin/base_site.html' %}
+{% load static %}
+
+{% block extrastyle %}
+  {{ block.super }}
+  {% if OPENWISP_TEST_MODE %}
+    <link rel="stylesheet" type="text/css" href="{% static "admin/css/openwisp.css" %}" />
+    <link rel="icon" type="image/x-icon" href="{% static "ui/openwisp/images/favicon.png" %}" />
```
pandafy, line 8: Is it required to explicitly add this? 
Shouldn't this be already present from `{{ block.super }}`?

tests/test_project/templates/admin/base_site.html
```diff
@@ -0,0 +1,16 @@
+{% extends 'admin/base_site.html' %}
+{% load static %}
+
+{% block extrastyle %}
+  {{ block.super }}
+  {% if OPENWISP_TEST_MODE %}
```
pandafy, line 6: Can we remove this variable too?

tests/test_project/templates/admin/base_site.html
niteshsinha17: During the test, those files are not loaded. That's why I added the new variable.

pandafy: These files are exactly the same, aren't they? 

https://github.com/openwisp/openwisp-utils/blob/9e182fd051e1b2142459014a9e5f6dc71a884ffa/tests/settings.py#L102-L120

niteshsinha17: I printed it value during test and result was ``[{'href': '/static/custom-admin-theme.css', 'rel': 'stylesheet', 'type': 'text/css', 'media': 'all'}]``. It is getting manipulated in some place.

pandafy: @niteshsinha17 this test are flawed.

https://github.com/openwisp/openwisp-utils/blob/9e182fd051e1b2142459014a9e5f6dc71a884ffa/tests/test_project/tests/test_admin.py#L203-L242

I was't aware of `mocks` back when I started contributing. Can you please fix these tests?

niteshsinha17: Thanks for the investigation. Fixed it.

tests/test_project/templates/admin/base_site.html
niteshsinha17: We need this variable to disable transitions during tests. But now it is not being served by utils.

pandafy: Any specific reason to not add follow css in [menu-test.css](https://github.com/openwisp/openwisp-utils/blob/master/tests/test_project/static/menu-test.css) file?

niteshsinha17: It will disable transition in the development environment too.

nemesisdesign:
Great work! Thank you very much @niteshsinha17 @atb00ker @pandafy :+1: :clap: :rocket:

PR #224, opened 2021-07-31
[enhancement] Show active menu link or group #216

### LINKED ISSUE #216: [gsoc21] Show active menu item/group
Screenshot from 2021-07-21 21-05-14

I am in the user page here but no indication is given in the menu. It would be appropriate to highlight the current active group or menu item.
If we can also expand the current group of which the item is part would be ideal.

niteshsinha17:
#### Changes

Added js to show current active menu link or group.
If the same link belongs to multiple groups then I am showing the first one as active.
#### Gif

activegroup

#### Checklist

closes #216

### DIFF
--- a/openwisp_utils/admin_theme/apps.py
+++ b/openwisp_utils/admin_theme/apps.py
@@ -25,5 +25,5 @@ class OpenWispAdminThemeConfig(AppConfig):
         # register dashboard item
         register_menu_group(
             position=10,
-            config={'label': _('Home'), 'url': '/admin/', 'icon': 'ow-dashboard-icon'},
+            config={'label': _('Home'), 'url': '/admin', 'icon': 'ow-dashboard-icon'},
         )
--- a/openwisp_utils/admin_theme/apps.py
+++ b/openwisp_utils/admin_theme/apps.py
@@ -25,5 +25,5 @@ class OpenWispAdminThemeConfig(AppConfig):
         # register dashboard item
         register_menu_group(
             position=10,
-            config={'label': _('Home'), 'url': '/admin', 'icon': 'ow-dashboard-icon'},
+            config={'label': _('Home'), 'url': '/admin/', 'icon': 'ow-dashboard-icon'},
         )
--- a/openwisp_utils/admin_theme/menu.py
+++ b/openwisp_utils/admin_theme/menu.py
@@ -98,8 +98,14 @@ class MenuLink(BaseMenuItem):
             raise ImproperlyConfigured(
                 f'"url" should be a type of "str". Error for the config- {config}'
             )
-        self.url = url
+        self.url = self.make_django_url(url)
         self.icon = config.get('icon')
+
+    def make_django_url(self, url):
+        # Add "/" at the end of link
+        if url[-1] != '/':
+            return url + '/'
+        return url
 
 
 class MenuGroup(BaseMenuItem):
--- a/openwisp_utils/admin_theme/static/admin/css/openwisp.css
+++ b/openwisp_utils/admin_theme/static/admin/css/openwisp.css
@@ -848,7 +848,9 @@ a.menu-item,
 .mg-head:focus,
 #menu .mg-link:focus,
 .select2-results__option--highlighted[aria-selected],
-.select2-results__option[aria-selected=true] {
+.select2-results__option[aria-selected=true],
+.active-menu-link,
+.active-mg .mg-head {
   background: #ffe5e5 !important;
   color: #df5d43 !important;
 }
@@ -874,7 +876,10 @@ a.menu-item,
 .mg-head:focus .mg-arrow,
 .menu-group.active .mg-head .icon,
 .mg-link:hover .icon,
-.menu-group.active .mg-arrow {
+.menu-group.active .mg-arrow,
+a.active-menu-link .icon,
+.active-mg .mg-head .icon,
+.active-mg .mg-head .mg-arrow {
   background: #df5d43;
 }
 .mg-arrow {
@@ -896,6 +901,9 @@ a.menu-item,
   overflow: hidden;
   transition: max-height 0.3s ease;
   background: #fdf2f2;
+}
+.mg-list li{
+  padding: 0;
 }
 
 /* for transition in the height of mg-dropdown */
--- a/openwisp_utils/admin_theme/static/admin/css/openwisp.css
+++ b/openwisp_utils/admin_theme/static/admin/css/openwisp.css
@@ -111,7 +111,7 @@ fieldset.module h2 {
 #main .inline-related h3 {
   padding: 15px;
   font-size: 15px;
-  background: #F9F9F9;
+  background: #f9f9f9;
 }
 td, th, li {
   font-size: 14px;
@@ -272,16 +272,16 @@ div.breadcrumbs {
 }
 #main .selector-chosen h2 {
   background: rgba(0, 0, 0, 0.6);
-  color: #fff ;
+  color: #fff;
 }
-span.datetimeshortcuts .date-icon{
+span.datetimeshortcuts .date-icon {
   mask: url(../../ui/openwisp/images/date.svg) no-repeat center;
   -webkit-mask: url(../../ui/openwisp/images/date.svg) no-repeat center;
   -webkit-mask-size: contain;
   -webkit-mask: contain;
   background: #df5d43;
 }
-span.datetimeshortcuts .clock-icon{
+span.datetimeshortcuts .clock-icon {
   mask: url(../../ui/openwisp/images/clock.svg) no-repeat center;
   -webkit-mask: url(../../ui/openwisp/images/clock.svg) no-repeat center;
   -webkit-mask-size: contain;
@@ -289,7 +289,7 @@ span.datetimeshortcuts .clock-icon{
   background: #df5d43;
 }
 span.datetimeshortcuts .date-icon:hover,
-span.datetimeshortcuts .clock-icon:hover{
+span.datetimeshortcuts .clock-icon:hover {
   background: #b31e00;
 }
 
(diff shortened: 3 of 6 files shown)


openwisp_utils/admin_theme/apps.py
```diff
@@ -25,5 +25,5 @@ def register_menu_groups(self):
         # register dashboard item
         register_menu_group(
             position=10,
-            config={'label': _('Home'), 'url': '/admin', 'icon': 'ow-dashboard-icon'},
+            config={'label': _('Home'), 'url': '/admin/', 'icon': 'ow-dashboard-icon'},
```
atb00ker, line 28: I hope the code isn't flaky enough that the presence of the last dash matters and not having it would break it.
~If it matters, it might be in our best interest to change the django side of code to use `urljoin` instead of appending strings to make the code more robust! :smile:
What do you think?~

openwisp_utils/admin_theme/static/admin/js/menu.js
```diff
@@ -207,3 +208,15 @@ function initToolTipHandlers() {
     });
   });
 }
+
+function showActiveItems(){
+  var activeLink = document.querySelector(`.nav a[href="${window.location.pathname}"]`);
```
atb00ker, line 213: Okay, I think this is why that `/` above was required.
From my point of view, why do we even need it?
Can't we add the active class to a button on click and remove the active class from all others. A simpler logic!
Do we really want highlighting when the users refreshes the page?

atb00ker (commented):
Tested the code, it works.
I have some questions about the approach:

openwisp_utils/admin_theme/apps.py
niteshsinha17: Yes, that slash matters. 
I think we need to change URL for the custom link. We can add ``/`` in the end of url provided by user only if it is missing.

openwisp_utils/admin_theme/static/admin/js/menu.js
niteshsinha17: I think we have to highlight group or item even on page refresh. @nemesisdesign What do you think?

nemesisdesign: the page always refreshes when a link is clicked guys... it's not a single page app

niteshsinha17: True 😂

nemesisdesign (changes requested):
Great progress @niteshsinha17!
Here's my feedback after testing:

- it would be great to open the menu group as well, so if we're in templates, show "configuration" highlighted and open, then "templates" highlighted as well
- it would be super great to highlight the links also when in one of the detail page, for example if we are in the details of a user, would be great to highlight the menu as if we are in the user list, is that possible @niteshsinha17?
- active menu items should have a different cursor, eg: `cursor: default`
- we could also style the active menu slightly differently to make it clear that the menu is active and not in hover state, peraphs we could play with the rgba version of the color and lower the opacity a bit?

niteshsinha17:
@thanks @nemesisdesign 
Here are the changes I have made according to your suggestions:
>it would be great to open the menu group as well, so if we're in templates, show "configuration" highlighted and open, then "templates" highlighted as well

I am disabling the group so that it remains always open. I will change the implementation if you want.

>it would be super great to highlight the links also when in one of the detail page, for example if we are in the details of a user, would be great to highlight the menu as if we are in the user list (in the narrow/minimized version we can highlight too but we shouldn't open the submenu), is that possible @niteshsinha17?
active menu items should have a different cursor, eg: cursor: default

:+1: Done

>we could also style the active menu slightly differently to make it clear that the menu is active and not in hover state, perhaps we could play with the rgba version of the color and lower the opacity a bit?

I have changed it but it is for testing purpose only( Please ignore it for some time). I will change it after approval.

openwisp_utils/admin_theme/static/admin/css/openwisp.css
```diff
@@ -953,6 +962,27 @@ h1#site-name {
   align-items: center;
 }
 
+/* Highlight menu item */
+
+a.active-menu-link
```
nemesisdesign, line 967: why not just `active`?

nemesisdesign (changes requested):
> @thanks @nemesisdesign
> Here are the changes I have made according to your suggestions:
> 
> > it would be great to open the menu group as well, so if we're in templates, show "configuration" highlighted and open, then "templates" highlighted as well
> 
> I am disabling the group too so that it remains always open. I will change the implementation if you want.

MM.. I'm not sure we need to be so strict, I'd still keep the normal behavior when clicking on other links.

> > it would be super great to highlight the links also when in one of the detail page, for example if we are in the details of a user, would be great to highlight the menu as if we are in the user list (in the narrow/minimized version we can highlight too but we shouldn't open the submenu), is that possible @niteshsinha17?
> > active menu items should have a different cursor, eg: cursor: default
> 
>  Done

Great!

> > we could also style the active menu slightly differently to make it clear that the menu is active and not in hover state, perhaps we could play with the rgba version of the color and lower the opacity a bit?
> 
> I have changed it but it is for testing purpose only( Please ignore it for some time). I will change it after approval.

Yea the functionality works so please go ahead 

Regarding the style, I did some testing with lowering the opacity but it does not convince me.
I think we can leave it as you did but I would simply make sure that hovering on an active menu group or menu item does not change its color.

openwisp_utils/admin_theme/static/admin/js/menu.js
atb00ker: Ah, I am getting spoiled by frameworks! 😄

tests/test_project/apps.py
```diff
@@ -136,3 +136,6 @@ def register_menu_groups(self):
         )
         register_menu_group(position=32, config=auth_config)
         register_menu_group(position=33, config=docs_config)
+        # TODO: Remove before merge
+        for i in range(20):
+            register_menu_group(position=34 + i, config=auth_config)
```
nemesisdesign, line 141: let's not forget this

nemesisdesign (changes requested):
Great work Nitesh, it looks really good and the fixed position + scrollbar also works very well.

I found one glitch, take a look:

glitch-active-menu

Clicking on any row in a list page closes the menu, that's weird, is this on purpose or is it a mistake?

niteshsinha17:
> Clicking on any row in a list page closes the menu, that's weird, is this on purpose or is it a mistake?

It's not a mistake. The menu group gets closed on clicking anywhere on the screen. Focus of the user also gets shifted from menu when they interact with other elements of the page. What do you think @nemesisdesign Do you want any change?

nemesisdesign:
> > Clicking on any row in a list page closes the menu, that's weird, is this on purpose or is it a mistake?
> 
> It's not a mistake. The menu group gets closed on clicking anywhere on the screen. Focus of the user also gets shifted from menu when they interact with other elements of the page. What do you think @nemesisdesign Do you want any change?

It's kind of weird to me, I think it may be also unnecessary and we may save some complexity here.
What do you think @atb00ker @pniaps @kkreitmair @nickberry17?

niteshsinha17:
> > > Clicking on any row in a list page closes the menu, that's weird, is this on purpose or is it a mistake?
> > 
> > 
> > It's not a mistake. The menu group gets closed on clicking anywhere on the screen. Focus of the user also gets shifted from menu when they interact with other elements of the page. What do you think @nemesisdesign Do you want any change?
> 
> It's kind of weird to me, I think it may be also unnecessary and we may save some complexity here.
> What do you think @atb00ker @pniaps @kkreitmair @nickberry17?

In any case when we need to close the menu group on clicking other elements when the menu is minimized. To not keep the same functionality on different screens we will need to introduces some other checks.

pniaps:
I think that the active menu has always tell me where I am, so  if I am in a sub item, the paren group should always be expanded, until y close it manually or navigate to another page.

kkreitmair:
I agree with @nemesisdesign  and @pniaps, because this will improve the workflow.

For example, if you do some configuration, you will mostly navigate through sub items of the configuration group. It will be annoying, to always first expand the group again, to get to the sub items.

nemesisdesign (commented):
> > > > Clicking on any row in a list page closes the menu, that's weird, is this on purpose or is it a mistake?
> > > 
> > > 
> > > It's not a mistake. The menu group gets closed on clicking anywhere on the screen. Focus of the user also gets shifted from menu when they interact with other elements of the page. What do you think @nemesisdesign Do you want any change?
> > 
> > 
> > It's kind of weird to me, I think it may be also unnecessary and we may save some complexity here.
> > What do you think @atb00ker @pniaps @kkreitmair @nickberry17?
> 
> In any case when we need to close the menu group on clicking other elements when the menu is minimized. To not keep the same functionality on different screens we will need to introduce some other checks.

Can you just do nothing and let the menu open the right section when the new page is loaded?

niteshsinha17:
Implemented requested feature.

openwisp_utils/admin_theme/menu.py
```diff
@@ -98,9 +98,15 @@ def __init__(self, config):
...
             )
-        self.url = url
+        self.url = self.make_django_url(url)
         self.icon = config.get('icon')
 
+    def make_django_url(self, url):
+        # Add "/" at the end of link
+        if url[-1] != '/':
+            return url + '/'
+        return url
```
nemesisdesign, line 108: not sure why this is needed, can you explain? May be good to add a comment.

nemesisdesign (approved):

openwisp_utils/admin_theme/menu.py
niteshsinha17: That is need because we compare url for links and current URL of page to find active link

PR #226, opened 2021-08-03
[fix] Fixed js not rendering issue in accounts page

niteshsinha17:
This PR is to solve all the issues that we will face during testing of the main menu.

###Change
Js file was not rendering because of the old check.

### DIFF
--- a/openwisp_utils/admin_theme/templates/admin/base_site.html
+++ b/openwisp_utils/admin_theme/templates/admin/base_site.html
@@ -1,4 +1,4 @@
-{% extends "base.html" %}
+{% extends "./base.html" %}
 {% load static i18n %}
 
 {% block title %}{{ title }} | {{ site_title }}{% endblock %}
@@ -65,7 +65,7 @@
     if we put it in extrahead it will be
     loaded before jQuery is loaded
 {% endcomment %}
-{% if has_permission %}
+{% if user.is_authenticated %}
     <script src="{% static "admin/js/menu.js" %}"></script>
 {% endif %}
 {% if OPENWISP_ADMIN_THEME_JS %}


nemesisdesign (changes requested):
I am testing this with https://github.com/openwisp/openwisp-users/pull/265 and I see a few problems.

Screenshot from 2021-08-03 18-22-32



- Spacing between paragraphs is inconsistent
- There's too much white space between the breadcrumbs and the content

I tried adding:

```css
#content p {
  margin: 10px 0
}
```

And tried removing:

```css
.title-wrapper h1,
.title-wrapper h2{
  float: left;
  margin-top: 20px!important;
  margin-bottom: 0px;
} 
```

And it looks better, but this obviously messes up other things in the admin, so I wonder, why is it necessary to make these titles float?

Can you please find a solution that works in both cases?

niteshsinha17:
@nemesisdesign Fixed it. I will be testing more today and will update you in chat we everything looks fine

openwisp_utils/admin_theme/static/admin/js/menu.js
```diff
@@ -14,8 +14,17 @@ var MenuTransitionTime = '0.1s';
...
 })();
 
+function cleanUp(){
+  // remove dublicate of objects-tools
+  var dublicateObject = document.querySelector('#content-main .object-tools');
+  if(dublicateObject){
+    dublicateObject.remove();
+  }
+}
+
```
nemesisdesign, line 27: What is this?

nemesisdesign (commented):
What about:

```css
#content p {
  margin: 10px 0
}
```

Is it causing issues? I think it works pretty well on non admin pages.

See my comment below.

openwisp_utils/admin_theme/static/admin/js/menu.js
niteshsinha17: Do you remember the overlapping issue of buttons and text on the device detail page.
To fix this I changed the position of ``object-tools``  but still it was being created twice by the django so these lines are used to remove the second one which was causing a repetition of some ids.

https://github.com/openwisp/openwisp-utils/blob/185b72ea675908aca8d12d53eede3dc03ef1bf47/openwisp_utils/admin_theme/templates/admin/base.html#L75-L85

Earlier it was inside ``block content``

niteshsinha17:
> What about:
> 
> ```css
> #content p {
>   margin: 10px 0
> }
> ```
> 
> Is it causing issues? I think it works pretty well on non admin pages.
> 
> See my comment below.

Now I got it why you suggested this addition. I have added this.
@kkreitmair and @pniaps Fixed this issue too here

image
image

pniaps:
I think that background should not be transparent on hover.

TransparentHover

openwisp_utils/admin_theme/static/admin/js/menu.js
nemesisdesign: It seems to me we should solve the issue which is causing this to be duplicated rather than adding code which removes the duplicates. Why is it duplicated in the first place? I suggest to create an issue for this.

nemesisdesign (changes requested):
Some conflicts appeared aftere merging https://github.com/openwisp/openwisp-utils/pull/224.

openwisp_utils/admin_theme/static/admin/js/menu.js
nemesisdesign: ok :+1: 
https://github.com/openwisp/openwisp-utils/issues/235

openwisp_utils/admin_theme/templates/admin/base_site.html
```diff
@@ -65,7 +65,7 @@ <h1 id="site-name"><a href="{% url 'admin:index' %}">{{ site_header }}</a></h1>
     if we put it in extrahead it will be
     loaded before jQuery is loaded
 {% endcomment %}
-{% if has_permission %}
+{% if user.is_authenticated %}
```
nemesisdesign, line 68: I think we should check whether `user.is_staff` as well, otherwise the menu is shown also in non admin templates.

niteshsinha17: You are right!

nemesisdesign: Let's leave it now, I created a new issue for this: https://github.com/openwisp/openwisp-utils/issues/236

nemesisdesign (approved):

PR #243, opened 2021-08-16
[fix] Fixed menu can be shown in non admin template #236

### LINKED ISSUE #236: [bug] The menu can be shown also in non admin templates
If a user is logged in and is browsing any django template which is not part of the admin, the menu can be shown anyway.

We should find a way to avoid the menu being showed in templates that do not belong to the admin.

niteshsinha17:
#### Changes
Added ``is_staff`` check.

#### Checklist

Closes #236

### DIFF
--- a/openwisp_utils/admin_theme/templates/admin/base.html
+++ b/openwisp_utils/admin_theme/templates/admin/base.html
@@ -23,25 +23,25 @@
   data-admin-utc-offset="{% now "Z" %}">
 
 <!-- Container -->
-<div id="container" class="toggle-menu{% if not user.is_authenticated %} no-auth{% endif %}" >
-  <div class="menu-backdrop"></div>
+<div id="container" class="toggle-menu {% if not user.is_authenticated or not user.is_staff %}no-auth{% endif %}" >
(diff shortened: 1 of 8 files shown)


nemesisdesign (changes requested):
Cool, it works, but we need at least one test @niteshsinha17, otherwise we risk that somebody may break this in the future.

pandafy (commented):
I don't like that the current PR uses `staff` flag of a user to determine whether to show or hide the menu. E.g. check in below screenshot: 
Screenshot from 2021-12-22 16-16-41

it is due to the fact that the non-admin site templates are inheriting from admin templates (check [openwisp_users/accounts/templates/base,html](https://github.com/openwisp/openwisp-users/blob/master/openwisp_users/accounts/templates/account/base.html)). 

I think we might need to tweak this inheritance pattern. Instead of adding theme to the `admin/base_site.html` directly, we should add it using includes. This will make templates more modular and will not require inheriting `admin/base_site.html` for non-admin templates.

nemesisdesign (approved):
Thanks!

PR #245, opened 2021-08-18
[docs] Improved consistency of menu function description #244

### LINKED ISSUE #244: [docs] Improve consistency of menu functions description
@niteshsinha17 please look at the way the rest of the README is written, for example:

https://github.com/openwisp/openwisp-utils#register_dashboard_template

> Allows including a specific django template in the OpenWISP dashboard.
> 
> It is designed to allow the inclusion of the geographic map shipped by OpenWISP Monitoring but can be used to include any custom element in the dashboard.

Try to follow this way in the text you have added.

"It allows us to register different menu item or group at different position ..."

Would become:

"It allows to register different menu items or groups at different positions ..."

Then, "If you want to add a link which contains a custom url then you can use following syntax for config."

would become:

"To add a link which contains a custom URL the following syntax can be used:"

See what I mean?

You kind of follow the same style, except in some phrases in which you use "you" and "us" a lot but would be good to avoid this.

See also https://github.com/openwisp/openwisp-utils#admin-filters which I had revised before merging, compare it to the sections coming before, you should notice the difference.

Let's try to make it more consistent.

niteshsinha17:
#### Checklist

closes #244

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -472,7 +472,7 @@ template ``context_processors`` in ``settings.py`` as shown below.
 ``register_menu_group``
 ^^^^^^^^^^^^^^^^^^^^^^^
 
-It allows us to register different menu item or group at different position in the Main Navigation Menu.
+It allows to register different menu item or group at the specified position in the Main Navigation Menu.
 
 **Syntax:**
 
@@ -532,8 +532,8 @@ Code example:
 .. note::
     An ``ImproperlyConfigured`` exception is raised if a menu element is already registered at the same position.
 
-    An ``ImproperlyConfigured`` exception is raised also if proper configuration is not provided based on the different types of
-    menu elements that you can register. Different types of configurations will be discussed in the next sections.
+    An ``ImproperlyConfigured`` exception is raised if proper configuration is not provided based on the different types of
+    menu elements that can be registered. Different types of configurations will be discussed in the next sections.
 
     It is recommended to use ``register_menu_group`` in the ``ready`` method of the ``AppConfig``.
 
@@ -544,7 +544,7 @@ Code example:
 Adding a generic link
 ~~~~~~~~~~~~~~~~~~~~~
 
-If you want to add a link which contains a custom url then you can use following syntax for config.
+To add a link which contains a custom url the following syntax can be used.
 
 **Syntax:**
 
@@ -572,15 +572,15 @@ Following is the description of the configuration:
 Adding a model link
 ~~~~~~~~~~~~~~~~~~~
 
-If you want to add a link that contains url of ``add`` or ``list`` page of a model
-then you can use following syntax. Users will only be able to see links for
+To add a link that contains url of ``add`` or ``list`` page of a model
+then the following syntax can be used. Users will only be able to see links for
 models they have permission to either view or edit.
 
 **Syntax:**
 
 .. code-block:: python
 
-    # adding a link of list page
+    # add a link of list page
     register_menu_group(
         position=1,
         config={
@@ -591,7 +591,7 @@ models they have permission to either view or edit.
         },
     )
 
-    # adding a link of add page
+    # add a link of add page
     register_menu_group(
         position=2,
         config={
@@ -621,7 +621,7 @@ Following is the description of the configuration:
 Adding a menu group
 ~~~~~~~~~~~~~~~~~~~
 
-If you want to add a nested menu group then you can use following syntax.
+To add a nested menu group of links the following syntax can be used.
 It creates a dropdown in the menu.
 
 **Syntax:**
@@ -663,7 +663,7 @@ Following is the description of the configuration:
 ``register_menu_subitem``
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
-It allows us to register an item in a registered group.
+It allows to register an item in a registered group.
 
 **Syntax:**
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -472,7 +472,7 @@ template ``context_processors`` in ``settings.py`` as shown below.
 ``register_menu_group``
 ^^^^^^^^^^^^^^^^^^^^^^^
 
-It allows us to register different menu item or group at different position in the Main Navigation Menu.
+It allows to register different menu item or group at the specified position in the Main Navigation Menu.
```
nemesisdesign, line 475: "to register different menu item", wouldn't it be better "to register a new menu item"?

README.rst
```diff
@@ -532,8 +532,8 @@ Code example:
 .. note::
     An ``ImproperlyConfigured`` exception is raised if a menu element is already registered at the same position.
 
-    An ``ImproperlyConfigured`` exception is raised also if proper configuration is not provided based on the different types of
-    menu elements that you can register. Different types of configurations will be discussed in the next sections.
+    An ``ImproperlyConfigured`` exception is raised if proper configuration is not provided based on the different types of
```
nemesisdesign, line 535: "if proper configuration is not provided", what about "if the supplied configuration is incorrect" or something similar?

README.rst
```diff
@@ -544,7 +544,7 @@ Code example:
 Adding a generic link
 ~~~~~~~~~~~~~~~~~~~~~
 
-If you want to add a link which contains a custom url then you can use following syntax for config.
+To add a link which contains a custom url the following syntax can be used.
```
nemesisdesign, line 547: url > URL

README.rst
```diff
@@ -572,15 +572,15 @@ Following is the description of the configuration:
 Adding a model link
 ~~~~~~~~~~~~~~~~~~~
 
-If you want to add a link that contains url of ``add`` or ``list`` page of a model
-then you can use following syntax. Users will only be able to see links for
+To add a link that contains url of ``add`` or ``list`` page of a model
```
nemesisdesign, line 575: url > URL.

BTW add and list are not code samples so I don't think we should highlight them as such, I belive  we can say add form and change list page

README.rst
```diff
@@ -663,7 +663,7 @@ Following is the description of the configuration:
 ``register_menu_subitem``
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
-It allows us to register an item in a registered group.
+It allows to register an item in a registered group.
```
nemesisdesign, line 666: What about "It allows to add an item in a registered group."?

README.rst
```diff
@@ -714,12 +714,12 @@ Code example:
     An ``ImproperlyConfigured`` exception is raised if group already has an
     item registered at ``item_position``.
 
-    You can only register a model link or a generic link.
-    An ``ImproperlyConfigured`` exception is raised
-    if you will try to register a group in place of item.
+    Registeration of a model link or a generic link is only possible.
```
nemesisdesign, line 717: Registeration contains a typo, but also doesn't sound great.

What about something like "It is only possible to register links to specific models or generic links" (can you specify what a generic link is? Do you mean an arbitrary/custom URL? If that's the case, I would refer to this as custom URL).

nemesisdesign (changes requested):
Good progress, some more refinement can be made.

README.rst
niteshsinha17: @nemesisdesign what do you think about this ``An ``ImproperlyConfigured`` exception is raised if supplied configuration do not match with the different types of possible configurations available( different configurations will be discussed in the next section).``

nemesisdesign: sounds good!

README.rst
niteshsinha17: I have written generic because we have a subheading ``Adding a generic link``. It will be easier to relate.

nemesisdesign: > I have written generic because we have a subheading `Adding a generic link`. It will be easier to relate.

Ok, I would still spend a few more words to explain generic link means any URL the user wants, with any word, no strong opinion on this.

niteshsinha17: I think then custom sounds better

README.rst
```diff
@@ -714,7 +714,7 @@ Code example:
     An ``ImproperlyConfigured`` exception is raised if group already has an
     item registered at ``item_position``.
 
-    Registeration of a model link or a generic link is only possible.
+    It is only possible to register links to specific models or custom URL.
```
pniaps, line 717: "generic link" => "custom URL" should be changed everywhere, not only here.

nemesisdesign: right, thanks!

nemesisdesign (changes requested):
Found a typo: "possible configurations available( different configurations will be discussed in the next section)"

The white space of the first rounded parenthesis is mistaken.

May I ask you to read again the whole text and double check everything before I do it please?

README.rst
```diff
@@ -472,7 +472,7 @@ template ``context_processors`` in ``settings.py`` as shown below.
 ``register_menu_group``
 ^^^^^^^^^^^^^^^^^^^^^^^
 
-It allows us to register different menu item or group at different position in the Main Navigation Menu.
+Allows registering a new menu item or group at the specified position in the Main Navigation Menu.
```
niteshsinha17, line 475: I have made this change to make it consistent with https://github.com/openwisp/openwisp-utils#register_dashboard_template

nemesisdesign (approved):
Thanks :+1:

PR #252, opened 2021-08-27
[feature] Added HelpTextStackedInline admin class

pandafy:
HelpTextStackedInline can be used to provide a help text for an
entire inline object

#### Changes

HelpTextStackedInline can be used to provide a help text for an
entire inline object

#### Screenshots

help-text-stacked-inline

#### Checklist

> #Closes #(issue-number)#


README.rst
```diff
@@ -831,6 +831,40 @@ An admin class that provides the UUID of the object as a read-only input field
...
+            'documentation_url': (
+                'https://github.com/openwisp/openwisp-utils'
+            )
+            # (optional) Icon to be shown along with help text. By default it uses
+            # "/static/admin/img/icon-alert.svg"
+            'image_url': '/static/admin/img/icon-alert.svg'
+        }
+
+.. figure:: https://github.com/openwisp/openwisp-utils/raw/media/docs/help-text-stacked-inline.png
+  :align: center
```
nemesisdesign, line 866: I would show this right after the heading. Don't you think it's better that way?

openwisp_utils/admin_theme/templates/admin/edit_inline/help_text_stacked.html
```diff
@@ -0,0 +1,47 @@
...
+    {% if inline_admin_formset.formset.can_delete and inline_admin_formset.has_delete_permission and inline_admin_form.original %}<span class="delete">{{ inline_admin_form.deletion_field.field }} {{ inline_admin_form.deletion_field.label_tag }}</span>{% endif %}
+  </h3>
+  {% comment %} OpenWISP customization begin {% endcomment %}
+  {% if inline_admin_formset.formset.help_text %}
+    {% with inline_admin_formset.formset.help_text as help_text %}
+        <div class="ow-help-text">
+            <img src="{{ help_text.image_url|default:'/static/admin/img/icon-alert.svg' }}">
+            <p>{{ help_text.text}}
+            {% if help_text.documentation_url %}
+                {% trans 'Please refer to the' %} <a href="{{help_text.documentation_url}}">{% trans 'documentation' %}</a> {% trans 'for more information' %}</p>
```
nemesisdesign, line 33: There's a problem here, we cannot split the phrase in 3 different translations, because the order of words changes from language to language.

I think we should use a single [blocktranslate tag](https://docs.djangoproject.com/en/3.2/topics/i18n/translation/#blocktranslate-template-tag) for this and pass the HTML tags as variables (eg: `open_link`, `close_link`).

nemesisdesign (changes requested):
Almost there! See below.

tests/test_project/admin.py
```diff
@@ -36,7 +37,7 @@ class OperatorForm(AlwaysHasChangedMixin, ModelForm):
     pass
 
 
-class OperatorInline(admin.StackedInline):
+class OperatorInline(HelpTextStackedInline):
     model = Operator
     form = OperatorForm
```
nemesisdesign, line 42: can you add `help_text` here?

nemesisdesign (changes requested):
Can you please add a test which checks for the presence of the `help_text`?

nemesisdesign (approved):
Thanks!

PR #266, opened 2021-12-02
[fix] Fix admin_theme links not getting compressed #237

### LINKED ISSUE #237: [bug] OPENWISP_ADMIN_THEME_LINKS does not play well with CompressStaticFilesStorage
I have the impression that static files defined in OPENWISP_ADMIN_THEME_LINKS do not get compressed by CompressStaticFilesStorage, I hope we can fix this.

devkapilbansal:
#### Changes

Minified files were not used in templates as the url is passed directly to `href` without processing. This PR fixes that.

#### Checklist

Closes #237

### DIFF
--- a/openwisp_utils/admin_theme/apps.py
+++ b/openwisp_utils/admin_theme/apps.py
@@ -1,4 +1,6 @@
 from django.apps import AppConfig
+from django.conf import settings
+from django.templatetags.static import static
 from django.utils.translation import gettext_lazy as _
 
 from . import settings as app_settings
@@ -13,6 +15,7 @@ class OpenWispAdminThemeConfig(AppConfig):
     def ready(self):
         admin_theme_settings_checks(self)
         self.register_menu_groups()
+        self.modify_admin_theme_settings_links()
         # monkey patch django.contrib.admin.apps.AdminConfig.default_site
         # in order to supply our customized admin site class
         # this is necessary in order to avoid having to modify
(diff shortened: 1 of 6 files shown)


openwisp_utils/admin_theme/apps.py
```diff
@@ -27,3 +30,15 @@ def register_menu_groups(self):
...
+        OPENWISP_ADMIN_THEME_LINKS = getattr(settings, 'OPENWISP_ADMIN_THEME_LINKS', [])
+        OPENWISP_ADMIN_THEME_JS = getattr(settings, 'OPENWISP_ADMIN_THEME_JS', [])
+
+        for css_files in OPENWISP_ADMIN_THEME_LINKS:
+            css_files['href'] = static(css_files['href'])
+
+        OPENWISP_ADMIN_THEME_JS = list(map(static, OPENWISP_ADMIN_THEME_JS))
+
+        setattr(settings, 'OPENWISP_ADMIN_THEME_LINKS', OPENWISP_ADMIN_THEME_LINKS)
+        setattr(settings, 'OPENWISP_ADMIN_THEME_JS', OPENWISP_ADMIN_THEME_JS)
```
nemesisdesign, line 44: it feels weird to do this here and overwrite the django setting.

What about doing this preprocessing here instead (by moving these settings as last and create a function which does the preprocessing)?
https://github.com/openwisp/openwisp-utils/blob/master/openwisp_utils/admin_theme/settings.py#L9-L10

tests/settings.py
```diff
@@ -105,19 +105,19 @@
 OPENWISP_ADMIN_THEME_LINKS = [
     {
         'type': 'text/css',
-        'href': '/static/admin/css/openwisp.css',
+        'href': 'admin/css/openwisp.css',
```
nemesisdesign, line 108: There's a good number of instances around which are deployed this way and it will be quite painful to update, so we must ensure backward compatibility.

Can you handle this in the preprocessing function? It should be as simple as using the string replace method, eg:

```python
static_file = static_file.replace('/static/, '')
```

nemesisdesign (changes requested):
Can you find a way to add a test for this functionality to ensure that if someone breaks it we find out please?

See also my comments below.

openwisp_utils/admin_theme/apps.py
devkapilbansal: I tried doing so but it will not work as apps are not loaded in beginning and `static` function will not work.

nemesisdesign (approved):
I pushed a number of improvements, please read the commit messages to understand why those changes were needed.

PR #281, opened 2022-02-14
[feature] Added custom celery task class #271

### LINKED ISSUE #271: [feature] Add celery task class for regular background tasks with default soft and hard time limit set
We should look into whether it's possible to create a celery task class in which we can set the hard and soft time limit to make sure tasks never become stale and once we achieve that we have to base all celery tasks (except the long running tasks which must in any case define some kind of soft or hard time limit) of all openwisp modules on this class.

pandafy:
OpenwispCeleryTask class allows to set soft and hard time limits for
celery tasks using django settings.

Closes #271

#### Changes

#### Screenshots

#### Checklist

> #Closes #(issue-number)#


openwisp_utils/tasks.py
```diff
@@ -0,0 +1,8 @@
+from celery import Task
+
+from . import settings as app_settings
+
+
+class OpenwispCeleryTask(Task):
+    soft_time_limit = app_settings.CELERY_SOFT_TIME_LIMIT
+    time_limit = app_settings.CELERY_HARD_TIME_LIMIT
```
pandafy, line 8: I wrote following tasks to test this 

```python
@shared_task(base=OpenwispCeleryTask)
def exceed_soft_time_limit():
    time.sleep(openwisp_utils_settings.CELERY_SOFT_TIME_LIMIT + 5)
```

And called it in following task

```python
def test_time_limits(self):
        tasks.exceed_soft_time_limit.delay()
        self.assertRaises(SoftTimeLimitExceeded, tasks.exceed_soft_time_limit.delay)
```

But the tests was not working even with eager mode.

README.rst
```diff
@@ -1006,6 +1006,27 @@ Extends ``collections.SortedDict`` and implements logic to sort inserted
...
 incurs memory space overhead.
 
+
+``openwisp_utils.tasks.OpenwispCeleryTask``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A custom celery task class that sets hard and soft time limits of celery tasks
+using `OPENWISP_CELERY_HARD_TIME_LIMIT <#openwisp_celery_hard_time_limit>`_
+and `OPENWISP_CELERY_SOFT_TIME_LIMIT <#openwisp_celery_soft_time_limit>`_
+settings respectively.
```
nemesisdesign, line 1016: let's make sure we explain that this task class should be used for regular background tasks but not for complex background tasks which can take a long time to execute (eg: firmware upgrades, network operations with retry mechanisms).

README.rst
```diff
@@ -1606,6 +1627,30 @@ This setting allows to change the logo which is displayed in HTML version of the
...
+Sets the soft time limit for celery tasks using
+`OpenwispCeleryTask <#openwisp_utilstasksopenwispcelerytask>`_.
+
+``OPENWISP_CELERY_HARD_TIME_LIMIT``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
++---------+---------------------+
+| type    | ``int``             |
++---------+---------------------+
+| default | ``30`` (in seconds) |
```
nemesisdesign, line 1648: wouldn't these defaults prevent using this class in many tasks we have now scattered around OW?

I'd be more conservative and use 30 seconds and 60 seconds.

nemesisdesign (changes requested):

README.rst
nemesisdesign: I think the hard time limit could even default to something higher, see https://github.com/openwisp/openwisp-notifications/pull/224#discussion_r807215214

nemesisdesign (commented):
PS: I've done some manual testing and it seems to work :+1.
The weird thing is that I expected some output in the celery worker which told the task had hit the time limit but this isn't showing up not even in debug mode, but I have verified code after the sleep is not executed.

nemesisdesign (approved):

nemesisdesign (approved):

PR #302, opened 2022-06-27
[fix] Fixed dashboard pie charts not showing 0 for no data #301

### LINKED ISSUE #301: [bug] Currently open wifi sessions does not show anything if zero
If there are zero wifi clients connected to the system I would expect the dashboard pie chart to show the number zero, but it doesn't show anything, just plain gray.

pandafy:
Fixes #301

--------------------------------
Screenshot from 2022-06-27 20-17-51
Screenshot from 2022-06-27 20-23-53


nemesisdesign (approved):
Great, thanks!

PR #303, opened 2022-06-30
[Feature] Added autocomplete filter #298

### LINKED ISSUE #298: [feature] Reusable autocomplete filter for DB relations
The goal of this issue is to take the logic written by the good @totallynotvaishnav in https://github.com/openwisp/openwisp-controller/pull/656, generalize it and make it reusable so that we can use this type of filter in all the OpenWISP modules which have admin filters pointing to database relations, which can get pretty big in production systems and hence need to be loaded with this tecnique (rather than loaded all at once). 

I believe that the first step is to address the issues I mentioned in https://github.com/openwisp/openwisp-controller/pull/656, then move that logic here to openwisp-utils, add basic tests and a dedicated README subsection section in the admin filters section: https://github.com/openwisp/openwisp-utils#admin-filters. We should also update the GIF if applicable.

codesankalp:
#### Changes

Added `AutocompleteFilter` in openwisp admin theme

#### Screenshots

auto_filter

#### Checklist

Closes #298

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -979,11 +979,11 @@ The derived filter class should define the ``queryset`` method as shown in follo
 ``openwisp_utils.admin_theme.filters.AutocompleteFilter``
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-The ``admin_theme`` sub app of this package provides an auto complete filter that uses
-django autocomplete widget to load filter data asynchronously.
+The ``admin_theme`` sub app of this package provides an auto complete
+filter that uses django autocomplete widget to load filter data asynchronously.
 
-This filter can be helpful when the number of objects is too large to load all at once
-which may cause the slow loading of the page.
+This filter can be helpful when the number of objects is too large
+to load all at once which may cause the slow loading of the page.
 
 .. code-block:: python
 
@@ -1002,13 +1002,14 @@ which may cause the slow loading of the page.
             MyAutoCompleteFilter,
             ...
         ]
-    
+
     @admin.register(MyOtherModel)
     class MyOtherModelAdmin(admin.ModelAdmin):
         search_fields = ['id']
 
-To customize or know more about it, please refer `django-admin-autocomplete-filter <https://github.com/farhan0581/django-admin-autocomplete-filter#usage>`_
-docs.
+To customize or know more about it, please refer to the
+`django-admin-autocomplete-filter documentation
+<https://github.com/farhan0581/django-admin-autocomplete-filter#usage>`_.
 
 Code utilities
 --------------
--- a/README.rst
+++ b/README.rst
@@ -114,6 +114,8 @@ Add ``openwisp_utils.admin_theme`` to ``INSTALLED_APPS`` in ``settings.py``:
         'django.contrib.staticfiles',
 
         'openwisp_utils.admin_theme',    # <----- add this
+        # add when using autocomplete filter
+        'admin_auto_filters',    # <----- add this
 
         'django.contrib.sites',
         # admin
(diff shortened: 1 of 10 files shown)


README.rst
```diff
@@ -974,6 +974,40 @@ The derived filter class should define the ``queryset`` method as shown in follo
             ...
         ]
 
+``openwisp_utils.admin_theme.filters.AutocompleteFilter``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
nemesisdesign, line 978:
```suggestion
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

tests/settings.py
```diff
@@ -22,6 +22,7 @@
     'django.contrib.sites',
     # admin
     'django.contrib.admin',
+    'admin_auto_filters',
```
nemesisdesign, line 25: let's not forget to update https://github.com/openwisp/openwisp-utils#using-the-admin_theme

nemesisdesign (changes requested):
Great progress @codesankalp!

I tested this on OpenWISP Controller (https://github.com/openwisp/openwisp-controller/pull/656/) and is working.

I have noticed the detail you pointed out regarding selecting the value and applying the filters right away and I think this behavior is inconsistent with what we have now so I want to look deeper into this before proceeding.

Can you make this code play well with the apply filter button defined here?

https://github.com/openwisp/openwisp-utils/blob/dd8eb5ec88ae8b228284523f33daabb8975b7f63/openwisp_utils/admin_theme/static/admin/js/ow-filter.js#L174-L195

See how that function looks for `.filter-options .selected` and then uses it to apply the filter?
I think we have to create a dummy element, hidden, which the existing JS filter code can find, to make it play nice.

Infact if I do this manually it works, that is, I add the following code after the hidden select and click apply the filter works fine:

```
<div class="filter-options">
  <a class="selected" href="?config__templates=<TEMPLATE_ID>"></a>
</div>
```

See the gif below for a demo:

autocomplete-filter

nemesisdesign (changes requested):
Great progress @codesankalp!

During testing I found out a few things which I think we should improve, this is a great feature which I want to use in different places, but before being able to do so we should handle some corner cases.

I tried creating filters in the device admin of OpenWISP Controller also for `group` and `organization`, mainly for testing purposes.

The findings for the organization field I resumed here: https://github.com/openwisp/openwisp-users/issues/270, it would be great to work on a prototype of that feature so that we can find out if we need to improve anything here in the base class to make that possible.

## Multiple filters bug

autocomplete-filter-multiple

## "All" label

You see that normal filters have the "All" label when no value is specified, while the autocomplete filter is blank?

Screenshot from 2022-07-08 11-52-52

Is it possible to make the autocomplete filter consistent and show "All" when no value is specified?

## Width of field

Screenshot from 2022-07-08 11-54-11

I see that the case of a long string in the options is handled pretty well by our theme already, so I wonder why is the autocomplete field so large compared to the other fields?

Can we make it of the same width?

openwisp_utils/admin_theme/static/admin/js/ow-auto-filter.js
```diff
@@ -2,15 +2,17 @@
 django.jQuery(document).ready(function () {
   // unbinding default event handlers of admin_auto_filters
   django.jQuery("#changelist-filter select, #grp-filters select").off("change");
-
+  django.jQuery(".auto-filter .select2-selection__placeholder").text("All");
```
nemesisdesign, line 5: `gettext("All")`

Probably this has to be called also after the filter selection is cleared (make a reusable function in that case please).

nemesisdesign (changes requested):
Good progress @codesankalp, I found a bit more quirks during testing.

## All label not shown when clearing filter

Try selecting an option, then click on the "x" icon in the filter to clear the selection, the "All" label is not shown:

Screenshot from 2022-07-11 19-21-56

## Interaction with multiple filters when clearing

I am testing this new filter on both templates and groups of the Device model in OpenWISP Controller, see the following GIF in which I perform the following steps:

- select a template in the template filter
- select a group in the group filter
- clear the group filter
- apply

Expected result: filter by template but not for group.
Actual result: it seems the group filter doesn't get cleared.

autocomplete-filter-quirks

nemesisdesign (approved):
:+1:

PR #316, opened 2022-09-29
[fix] Fixed autocomplete filter to filter objects with null value

pandafy:
#### Changes

#### Screenshots

#### Checklist

> #Closes #(issue-number)#


openwisp_utils/admin_theme/admin.py
```diff
@@ -28,6 +28,19 @@ def index(self, request, extra_context=None):
             context = {'dashboard_enabled': False}
         return super().index(request, extra_context=context)
 
+    def get_urls(self):
+        from django.urls import path
+
+        from .views import AutocompleteJsonView
```
nemesisdesign, line 34: it seems to me that importing these two at the top shouldn't be an issue

pandafy:
> During testing with openwisp-monitoring, when I select the null option the URL becomes `http://localhost:8000/admin/config/device/?e=1`, can you try the same on openwisp-monitoring and let me know if you get the same result please?

@nemesisdesign did you install https://github.com/openwisp/openwisp-controller/pull/698 ? I was able to replicate this issue with older version of openwisp-controller. Can you please test after installing this PR?

nemesisdesign (approved):

PR #274, opened 2022-01-24
[enhancement] Added body color explicitly #272

Aryamanz29:
#### Changes

- Added style variables to prevent color changes in dark mode
Closes #272

#### BEFORE 📸 

Screenshot (239)

Screenshot (244)

#### AFTER 📸 

Screenshot (241)

Screenshot (246)

#### Checklist

- [x] I have read the [contributing guidelines](http://openwisp.io/docs/developer/contributing.html#how-to-commit-your-changes-properly).
- [x] I have manually tested the proposed changes.
- [ ] I have written new test cases to avoid regressions. (if necessary)
- [ ] I have updated the documentation. (e.g. README.rst)
- [ ] I have added [change!] to commit title to indicate a backward incompatible change. (if required)
- [ ] I have checked the links added / modified in the documentation.

PR #321, opened 2022-10-11
[feature] Added functions to register-unregister admin theme #274

pandafy:
Closes #274

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -212,13 +212,16 @@ to provide CSS and JS files to customise admin theme.
 
     You can learn more in the `Django documentation <https://docs.djangoproject.com/en/3.0/ref/settings/#std:setting-STATICFILES_DIRS>`_.
 
-Adding admin theme programmatically
+Extend admin theme programmatically
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 ``openwisp_utils.admin_theme.theme.register_theme_link``
 """"""""""""""""""""""""""""""""""""""""""""""""""""""""
 
 Allows adding items to `OPENWISP_ADMIN_THEME_LINKS <#openwisp_admin_theme_links>`_.
+
+This function is meant to be used by third party apps or OpenWISP modules which
+aim to extend the core look and feel of the OpenWISP theme (eg: add new menu icons).
 
 **Syntax:**
 
@@ -237,6 +240,9 @@ Allows adding items to `OPENWISP_ADMIN_THEME_LINKS <#openwisp_admin_theme_links>
 """"""""""""""""""""""""""""""""""""""""""""""""""""""""""
 
 Allows removing items from `OPENWISP_ADMIN_THEME_LINKS <#openwisp_admin_theme_links>`_.
+
+This function is meant to be used by third party apps or OpenWISP modules which
+aim additional functionalities to UI of OpenWISP (eg: adding a support chatbot).
 
 **Syntax:**
 
--- a/README.rst
+++ b/README.rst
@@ -183,7 +183,7 @@ Supplying custom CSS and JS for the admin theme
 Add ``openwisp_utils.admin_theme.context_processor.admin_theme_settings`` to
 template ``context_processors`` in ``settings.py`` as shown below.
 This will allow to set `OPENWISP_ADMIN_THEME_LINKS <#openwisp_admin_theme_links>`_
-and `OPENWISP_ADMIN_THEME_JS <openwisp_admin_theme_js>`_ settings
+and `OPENWISP_ADMIN_THEME_JS <#openwisp_admin_theme_js>`__ settings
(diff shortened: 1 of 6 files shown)


README.rst
```diff
@@ -183,7 +183,7 @@ Supplying custom CSS and JS for the admin theme
 Add ``openwisp_utils.admin_theme.context_processor.admin_theme_settings`` to
 template ``context_processors`` in ``settings.py`` as shown below.
 This will allow to set `OPENWISP_ADMIN_THEME_LINKS <#openwisp_admin_theme_links>`_
-and `OPENWISP_ADMIN_THEME_JS <openwisp_admin_theme_js>`_ settings
+and `OPENWISP_ADMIN_THEME_JS <#openwisp_admin_theme_js>`__ settings
```
nemesisdesign, line 186: Thanks for fixing this :+1:

README.rst
```diff
@@ -212,6 +212,81 @@ to provide CSS and JS files to customise admin theme.
 
     You can learn more in the `Django documentation <https://docs.djangoproject.com/en/3.0/ref/settings/#std:setting-STATICFILES_DIRS>`_.
 
+Adding admin theme programmatically
```
nemesisdesign, line 215:
```suggestion
Extend admin theme programmatically
```

README.rst
```diff
@@ -212,6 +212,81 @@ to provide CSS and JS files to customise admin theme.
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+``openwisp_utils.admin_theme.theme.register_theme_link``
+""""""""""""""""""""""""""""""""""""""""""""""""""""""""
+
+Allows adding items to `OPENWISP_ADMIN_THEME_LINKS <#openwisp_admin_theme_links>`_.
```
nemesisdesign, line 221:
```suggestion
Allows adding items to `OPENWISP_ADMIN_THEME_LINKS <#openwisp_admin_theme_links>`_.

This function is meant to be used by third party apps or OpenWISP modules which aim extend the core look and feel of the OpenWISP theme (eg: add new menu icons).
```

openwisp_utils/admin_theme/theme.py
```diff
@@ -0,0 +1,66 @@
...
+    if not isinstance(links, list):
+        raise ImproperlyConfigured(
+            '"openwisp_utils.admin_theme.theme.register_theme_link"'
+            ' accepts "list" of links.'
+        )
+    # We don't raise ImproperlyConfigured exception if
+    # link is already present in THEME_LINKS because
+    # user might specify a link in project settings which is
+    # already registered by an application.
+    # This would lead to time-consuming debugging for little returns.
```
nemesisdesign, line 19: Are you sure this is consistent with what we do across all the openwisp codebase?

If I am not mistaken, in these functions which register new items in a data structure, we usually let it fail early and complaining loud if there's any duplicate registration attempted (which is a good thing and a best practice in the Python community to fail early and loud).

It doesn't look consistent with what we've been doing up to now to me.

openwisp_utils/admin_theme/theme.py
```diff
@@ -0,0 +1,66 @@
...
+    if not isinstance(jss, list):
+        raise ImproperlyConfigured(
+            '"openwisp_utils.admin_theme.theme.register_theme_js"'
+            ' accepts "list" of JS.'
+        )
+    # We don't raise ImproperlyConfigured exception if
+    # js is already present in THEME_JS because
+    # user might specify a js in project settings which is
+    # already registered by an application.
+    # This would lead to time-consuming debugging for little returns.
```
nemesisdesign, line 50: same here

openwisp_utils/admin_theme/theme.py
```diff
@@ -0,0 +1,66 @@
...
+            '"openwisp_utils.admin_theme.theme.unregister_theme_link"'
+            ' accepts "list" of links.'
+        )
+    for link in links:
+        try:
+            THEME_LINKS.remove(link)
+        except ValueError:
+            raise ImproperlyConfigured(
+                f'{link["href"]} was not added to OPENWISP_ADMIN_THEME_LINKS'
+            )
```
nemesisdesign, line 37: This is good but also inconsistent with what's being done in the register function: the register won't complain if there's a duplicate but the unregister will complain if there's nothing to remove, the behavior is not consistent even within the same file. We have to avoid adding inconsistencies across the code base, let's decide one way and stick to it. I believe that failing early and loud if a piece of code is badly written (tries to register an item which has been already registered or tries to remove an item which is not present) is better because it allows developers to find out mistakes early instead of later on when they least expect it.

nemesisdesign (changes requested):
Looks good but I see areas for improvement in consistency and docs, please see my comments below.

nemesisdesign (approved):
This looks good, let's wait for a bit more testing before merging.

PR #334, opened 2023-03-06
[chores] Fixed no icon and added generic help icon

nemesisdesign:
Fixed UI when icon is not specified and added a generic help icon.


nemesisdesign:
Merged manually (mistake).

PR #336, opened 2023-03-23
[fix] Fixed CopyableFieldsAdmin fields order #335

### LINKED ISSUE #335: [bug] CopyableFieldsAdmin.get_fields() modifies the order of fields
The change introduced in https://github.com/openwisp/openwisp-utils/commit/b10d3cdd1f3eadf3d224ccb5f078221b181f6721 has an intended side effect, the order of the fields is altered, which is something we do not want.

Here's the result in the add organization form:

Screenshot from 2023-03-22 20-31-01

Aryamanz29:
#### Before

Screenshot from 2023-03-23 16-56-27

#### After

Screenshot from 2023-03-23 16-56-40

Fixes #335


nemesisdesign (changes requested):
Looks good but we are missing a test for this bug.

I see the same issue in `ProjectAdmin` in the test project of this module, I think a possible test would be to instantiate `ProjectAdmin` for a new object (like opening the add form page), call `get_fields` and `get_readonly_fields` and expect the order to match what is defined in the code. I expect this test to fail without the patch.
The test should be added to `test_project.tests.test_admin`.

Without the patch:

Screenshot from 2023-03-23 14-10-52

With the patch:

Screenshot from 2023-03-23 14-10-43

Aryamanz29:
I found that it's the conversion of a tuple to a set or vice versa that causes the random admin field order @nemesisdesign.


### Without patch
 
- When [we convert a tuple to a set in python](https://github.com/openwisp/openwisp-utils/pull/336/files#diff-0b0aa7023ab815c11b06105a1dd00e47cb5589f9c26b37786903fa5885c0a710L107), the order of the elements in the tuple is not preserved because sets do not have an inherent order. Sets are implemented using a hash table, which means that the elements are stored in an arbitrary order based on their hash values. This order can change each time the set is created or modified, that's why I removed this in [`cb78d18` (#336)](https://github.com/openwisp/openwisp-utils/pull/336/commits/cb78d18516e2bca54066ce9ae304d7a396cde68e)

- When running the below sample test (`test_copyablefields_admin_fields_order`) a few times, it gives me a different result each time:

```py
    def test_copyablefields_admin_fields_order(self):
        path = reverse('admin:test_project_project_add')
        response = self.client.get(path)
        ma = ProjectAdmin(Project, self.site)
        # uuid should be missing in ma.get_fields  
        # because we're testing project admin add form
        print(ma.fields, ma.get_fields(self.client.request))
        print(ma.readonly_fields, ma.get_readonly_fields(self.client.request))   
```

#### Results:

You may have noticed that we are receiving random admin field orders, as shown below. Therefore, if you remove this patch and run the new test case that I recently added in [`5deafdb` (#336)](https://github.com/openwisp/openwisp-utils/pull/336/commits/5deafdb81d0c378dad048be9801499952a0a1988), it is possible that the test might pass instead of failing. This is because the set conversion behavior, which I already explained above.

Screenshot from 2023-03-24 00-17-05

nemesisdesign (approved):
All is clear, thanks @Aryamanz29.

Working with data structures to represent application data is not trivial and can sometimes lead to unintended consequences like this one. Let's keep it in mind for the future.

PR #349, opened 2023-06-30
[chores] Fixed changes in theme due to Django 4.2.0

pandafy:
- Fixed receive URL field
- Fixed form row labels


nemesifier (changes requested):
Can you please provide a screenshot of the UI quirks?
That way I can not only verify that the problems are fixed but also test on the older version of Django.

pandafy:
### Receive URL field 

#### On Django 4.2  without fixes 

Screenshot from 2023-07-03 20-19-38

#### On Django 4.2 with fixes 

Screenshot from 2023-07-03 20-20-00

### Label CSS fix  

This primarily affects the custom generated fields in configuration editor. But, I have seen it's affect at other places in the admin (e.g. alert setting inline admin on device change page)

#### On Django 4.2 without fixes 
Screenshot from 2023-07-03 20-31-16

Screenshot from 2023-07-03 20-26-46

#### On Django 4.2 with fixes 

Screenshot from 2023-07-03 20-30-47

Screenshot from 2023-07-03 20-32-36

nemesifier (approved):

PR #365, opened 2023-11-29
[feature] Added optional clean insights measurement collection #360

### LINKED ISSUE #360: [feature] Add optional clean insights data collection
Depends on https://github.com/openwisp/openwisp-utils/issues/273

- schema:[https://gitlab.com/cleaninsights/clean-insights-design/-/blob/master/schema-docs/README.md](https://gitlab.com/cleaninsights/clean-insights-design/-/blob/master/schema-docs/README.md "smartCard-inline") - we should focus on events
- info to collect:
  - openwisp modules enabled
  - version of the modules
  - OS version
  - available number of CPUs, total RAM
- in order to know if the data changes over time (needed to track upgrades), we need to store some data in the DB

We can collect this info when the application starts, maybe we should restrict this to the application server process only (no celery) and ensure it's performed only once, we should use a thread to avoid slowing down startup. 

We need a setting to control whether this is enabled or not, it shall be disabled by default.

Let's add a section in the README which explains what data is collected and why.

We can test this in staging with the clean insights staging/prod system.

pandafy:
Closes #360

#### Changes

#### Screenshots

#### Checklist

> #Closes #(issue-number)#


openwisp_utils/measurements/tasks.py
```diff
@@ -0,0 +1,67 @@
...
+                    break
+            else:
+                message = str(error)
+            logger.warning(
+                f'Error posting clean insights events: {message}. Retrying in 5 seconds.'
+            )
+            time.sleep(5)
+    logger.error(
+        f'Maximum tries reach to upload Clean Insights measurements. Error: {message}'
+    )
```
nemesifier, line 50: Let's use the retry mechanism offered by the requests library that you know of ;)

README.rst
```diff
@@ -1615,6 +1616,20 @@ This backend extends ``django.contrib.gis.db.backends.spatialite``
...
 
+Clean Insights Measurements
+---------------------------
+
+openwisp-utils includes an optional sub-app ``openwisp_utils.measurements``.
+This sub-app enables collection of following measurements:
+
+- OpenWISP Version
+- Enabled OpenWISP modules and their versions
+- OS details
```
nemesifier, line 1627: - Whether the event is related to a new installation or an upgrade

nemesifier: I think this still needs to be updated.

README.rst
```diff
@@ -54,6 +54,7 @@ Current features
 * `Admin Theme utilities <#admin-theme-utilities>`_
 * `REST API utilities <#rest-api-utilities>`_
 * `Test utilities <#test-utilities>`_
+* `Clean Insights Measurements <#clean-insights-measurements>`_
```
nemesifier, line 57: Let's rename this section as `Collection of Usage Metrics`.

README.rst
```diff
@@ -1615,6 +1616,21 @@ This backend extends ``django.contrib.gis.db.backends.spatialite``
 database backend to implement a workaround for handling
 `issue with sqlite 3.36 and spatialite 5 <https://code.djangoproject.com/ticket/32935>`_.
 
+Clean Insights Measurements
+---------------------------
+
+openwisp-utils includes an optional sub-app ``openwisp_utils.measurements``.
```
nemesifier, line 1622:
```suggestion
The openwisp-utils module includes an optional sub-app ``openwisp_utils.measurements``.
```

README.rst
```diff
@@ -1615,6 +1616,21 @@ This backend extends ``django.contrib.gis.db.backends.spatialite``
...
+This sub-app enables collection of following measurements:
+
+- OpenWISP Version
+- Enabled OpenWISP modules and their versions
+- OS details
+- Whether the event is related to a new installation or an upgrade
+
+We collect data on OpenWISP usage to gauge user engagement, satisfaction,
+and upgrade patterns. This informs our development decisions, ensuring
+continuous improvement aligned with user needs.
```
nemesifier, line 1632: Let's add a paragraph introducing and linking to cleaninsights.

openwisp_utils/measurements/tasks.py
```diff
@@ -0,0 +1,55 @@
...
+    get_openwisp_version,
+    get_os_details,
+)
+
+from ..utils import retryable_request
+from .models import OpenwispVersion
+from .utils import _get_events, get_openwisp_module_events, get_os_detail_events
+
+CLEAN_INSIGHTS_URL = 'https://analytics.openwisp.io/cleaninsights.php'
+MAX_TRIES = 3
```
nemesifier, line 15: what about setting this to 10? I don't see this as being used anywhere BTW. Are you sure it's not a leftover?

openwisp_utils/measurements/tasks.py
```diff
@@ -0,0 +1,55 @@
...
+            },
+        )
+        assert response.status_code == 204
+    except Exception as error:
+        if isinstance(error, AssertionError):
+            message = f'HTTP {response.status_code} Response'
+        else:
+            message = str(error)
+        logger.error(
+            f'Maximum tries reached to upload Clean Insights measurements. Error: {message}'
```
nemesifier, line 37:
```suggestion
            f'Collection of usage metrics failed, max retries exceeded. Error: {message}'
```

openwisp_utils/measurements/tasks.py
```diff
@@ -0,0 +1,55 @@
...
+        response = retryable_request(
+            'post',
+            url=CLEAN_INSIGHTS_URL,
+            json={
+                'idsite': 5,
+                'events': events,
+            },
+        )
+        assert response.status_code == 204
+    except Exception as error:
```
nemesifier, line 31: are you absolutely sure this code works also when the server does not reply at all?

nemesifier (changes requested):
Let's not mention clean insights in the code and comments except in the README.
Let's not assume people reading the code know what clean insights is, let's use a generic term "Usage Metric Collection".

See my comments below.

openwisp_utils/measurements/tasks.py
```diff
@@ -0,0 +1,55 @@
...
+def send_clean_insights_measurements(upgrade_only=False):
+    current_versions = get_enabled_openwisp_modules()
+    current_versions.update({'OpenWISP Version': get_openwisp_version()})
+    events = []
+    events.extend(get_os_detail_events(get_os_details()))
+    if OpenwispVersion.is_new_installation():
+        events.extend(_get_events('Install', current_versions))
+        OpenwispVersion.objects.create(module_version=current_versions)
+    else:
+        upgraded_modules = OpenwispVersion.get_upgraded_modules(current_versions)
```
pandafy, line 51: On a new installation, the database tables would not present and database operations would result in an error.

Shall we update this task to check whether the database tables are present?

pandafy: I am thinking to make the task retry itself if such an database error is encountered.

nemesifier (commented):
Can you look into why the code coverage is decreasing?

openwisp_utils/measurements/tasks.py
nemesifier: is this resolved?

openwisp_utils/measurements/tasks.py
pandafy: I have tested the following scenarios: 

1. Trying to establish connection to the port where no application is listening - Check the [command output on Hastebin](https://hastebin.skyra.pw/ivuniyihat.py)
2. When the server time outs (hangs) after establishing the connection - Check the [command output on Hastebin](https://hastebin.skyra.pw/jeyowecehe.py)

I have added a test case for this https://github.com/openwisp/openwisp-utils/pull/365/files#diff-118b67d11be9c05a77bccfedd814f41974ca3fb2bfd21eb010350faf22d59d3bR180-R196

openwisp_utils/measurements/tasks.py
pandafy: This error would only be occur if someone manually triggers the task before the database migrations are run. 

Since the signal receiver is executing at `post_migrate`, it means that the database migrations have been already executed.

openwisp_utils/utils.py
```diff
@@ -62,3 +65,17 @@ def print_color(string, color_name, end='\n'):
...
     color = color_dict.get(color_name, '0')
     print(f'\033[{color}m{string}\033[0m', end=end)
+
+
+def retryable_request(method, timeout=(4, 8), max_retries=3, **kwargs):
+    request_session = requests.Session()
+    retries = Retry(
+        total=max_retries,
+        backoff_factor=1,
+        status_forcelist=[429, 500, 502, 503, 504],
```
nemesifier, line 75: This new utility function will be useful, please allow to custoimize the arguments passed to `Retry()`, eg:

- backoff_factor
- backoff_jitter
- status_forcelist
- allowed_methods
- add another argument called retry_kwargs=None

Then I recommend doing something  like:

```python
if retry_kwargs is None:
    retry_kwargs = {}
retry_kwargs.update(dict(
    total=max_retries,
    backoff_factor=backoff_factor,
    backoff_jitter=backoff_jitter,
    status_forcelist=status_forcelist,
    allowed_methods=allowed_methods
))
retries = Retry(**retry_kwargs)
```

This will allow us to reuse this function in other modules because as you know it's a common need.
Please add this function to the README and make sure there's at least one unit test dedicated to this function only.

nemesifier (changes requested):
I think the metric collection logic is good to merge, but I think the `retryable_request` needs a bit more love so we can reuse it in other OpenWISP python modules too.

openwisp_utils/utils.py
```diff
@@ -67,14 +67,39 @@ def print_color(string, color_name, end='\n'):
...
-        status_forcelist=[429, 500, 502, 503, 504],
-        allowed_methods=['DELETE', 'GET', 'HEAD', 'OPTIONS', 'PUT', 'TRACE', 'POST'],
+def retryable_request(
+    method,
+    timeout=(4, 8),
+    max_retries=3,
+    backoff_factor=1,
+    backoff_jitter=0.0,
+    status_forcelist=(429, 500, 502, 503, 504),
+    allowed_methods=('HEAD', 'GET', 'PUT', 'DELETE', 'OPTIONS', 'TRACE', 'POST'),
```
nemesifier, line 77: one more time, **let's not use mutable default arguments** for the reasons described here:
https://dev.to/sharmapacific/mutable-default-arguments-in-python-22ce

We have already been hit a few times by obscure bugs which originated from using mutable default arguments, let's not repeat the same mistakes.

This applies to timeout, status_forcelist, allowed_methods.

nemesifier (changes requested):

nemesifier (approved):
After giving a second look to the mutable argument subject, I think it's ok since we don't return the mutable objects nor manipulate these inside the function, so we are not risking to be hit by weird bugs and this usage is fine.

PR #373, opened 2024-03-14
[fix] Fixed several minor issues in retryable requests

nemesifier:
- we need to ensure a recent version of requests is installed
- docs had several mistakes
- some code was repeated twice in the function definition

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,6 +46,8 @@ setup(
         'django-compress-staticfiles~=1.0.1b',
         'django-admin-autocomplete-filter~=0.7.1',
         'swapper~=1.3.0',
+        # allow wider range here to avoid interfering with other modules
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -46,6 +46,8 @@
         'django-compress-staticfiles~=1.0.1b',
         'django-admin-autocomplete-filter~=0.7.1',
         'swapper~=1.3.0',
+        # allow wider range here to avoid interfering with other modules
+        'requests>=2.31.0,<3.0.0',
```
pandafy, line 50: Do we need an upper limit here?

nemesifier: just to avoid potential issues in the future, we can change it when the new version will be released

pandafy (approved):
Thank you these fixes @nemesifier, LGTM!

PR #378, opened 2024-04-15
[change] Metric collection: added consent in web UI #372

### LINKED ISSUE #372: [change] Metric collection: add consent in web UI
Improvements discussed recently:

- Add the consent screen to the system info page (we'll need to make sure only superusers can see/change this), we'll keep it enabled by default
- If necessary, move metric collection to one of the existing subapps (optional)

We need to find a way to notify users the first time they access the system. We could simply use the Django message framework and produce a message the first time a super user logs in or anything equivalent.

## More notes

- We want to show an info message in the admin, but we want to show it only once, the info message will contain a link to the system info page and inform the user they can opt out
- In the system info page we can have an additional template block which shows a consent checkbox which can be turned off (opt out)
- We probably need to change the index() view of `OpenwispAdminSite`
- It's  not a problem that `admin_theme` will be slightly aware of the measurements app

pandafy:
Closes #372

#### Changes

#### Screenshots

#### Checklist

> #Closes #(issue-number)#


openwisp_utils/measurements/models.py
```diff
@@ -53,3 +54,32 @@ def log_module_version_changes(cls, current_versions):
...
+    # Disabling it is a one-time action by the superuser. Whenever a superuser
+    # disables metric collection, they are opting-out to share anymore anonymous
+    # usage metrics with the OpenWISP project.
+    user_consented = models.BooleanField(
+        default=True,
+        verbose_name=_('Allow collecting anonymous usage metrics'),
+        help_text=_(
+            'Allow OpenWISP to collect and share anonymous usage metrics to improve'
+            ' the software. Before opting-out kindly consider reading'
+            ' <a href="https://github.com/openwisp/openwisp-utils?tab=readme-ov-file'
```
nemesifier, line 82: link openwisp.io/doc

openwisp_utils/admin_theme/templates/admin/openwisp_info.html
```diff
@@ -15,4 +15,29 @@ <h2>{% trans "OS Information" %}</h2>
 <p><strong>{% trans "OS version" %}:</strong> {{ system_info.os_version }}</p>
 <p><strong>{% trans "Kernel version" %}:</strong> {{ system_info.kernel_version }}</p>
 <p><strong>{% trans "Hardware platform" %}:</strong> {{ system_info.hardware_platform }}</p>
+
+{% block metric_collect_consent %}
+    {% if metric_collection_installed %}
+    <h2>Metric collection</h2>
+        <form method="POST" id="id_metric_collection_consent_form">
+            {% csrf_token %}
+            {{ metric_consent_form }}
```
nemesifier, line 24: Please rewrite this so that we can have the input first and each line wrapped on a `<p>` element, the current output is not readable.

You probably need to add:

```
    margin-right: 8px;
    vertical-align: top;
```
To the checkbox input to make it look decent:

Screenshot from 2024-04-16 11-52-42

nemesifier (changes requested):
See my changes to the README here: https://github.com/openwisp/openwisp-utils/commit/95c8e9a592fe68f9e73a5251ade27bb8e2e882a6, please mention the possibility of removing consent from the UI there too.

openwisp_utils/admin_theme/templates/admin/openwisp_info.html
pandafy: I just used some CSS to achieve the same result. Re-writing this would require copying some code from Django which we will need to maintain over time.

nemesifier (approved):
Looks pretty good to me now after these last iterations.
I am merging, if you spot anything else that needs improvement please send a new PR.

PR #423, opened 2024-12-18
[ci] Added GitHub workflow to replicate commits to version branch

pandafy:
## Checklist

### DIFF
--- a/.github/workflows/version-branch.yml
+++ b/.github/workflows/version-branch.yml
@@ -1,0 +1,12 @@
+name: Replicate Commits to Version Branch
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  version-branch:
+    uses: openwisp/openwisp-utils/.github/workflows/reusable-version-branch.yml@master
+    with:
+      module_name: openwisp_utils
--- a/openwisp_utils/__init__.py
+++ b/openwisp_utils/__init__.py
@@ -1,4 +1,4 @@
-VERSION = (1, 1, 1, 'final')
+VERSION = (1, 2, 0, 'alpha')
 __version__ = VERSION  # alias
 
 
--- a/.github/workflows/reusable-version-branch.yml
+++ b/.github/workflows/reusable-version-branch.yml
@@ -1,0 +1,51 @@
+name: Replicate Commits to Version Branch
+
+on:
+  workflow_call:
+    inputs:
+      module_name:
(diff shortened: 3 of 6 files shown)


pandafy:
I have tested the GitHub action on my fork: 

- [Creating a new branch from the version number](https://github.com/pandafy/openwisp-utils/actions/runs/12392581882) 
- [Replicating the master branch](https://github.com/pandafy/openwisp-utils/actions/runs/12392528667)

.github/workflows/version-branch.yml
```diff
@@ -0,0 +1,48 @@
...
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+
+    - name: Get version
+      id: get_version
+      run: |
+        VERSION=$(python -c '
+        from openwisp_utils import VERSION
+        print(f"{VERSION[0]}.{VERSION[1]}.x")
```
nemesifier, line 30: Wouldn't this produce something like `1.2.x`? We are using branch names without the x now.

.github/workflows/version-branch.yml
```diff
@@ -0,0 +1,48 @@
...
+    - name: Rebase changes onto version branch
+      run: |
+        if git ls-remote --heads origin $VERSION | grep -sw $VERSION; then
+          git fetch origin --unshallow
+          git checkout -b $VERSION origin/$VERSION
+          git rebase origin/master
+        else
+          git checkout -b $VERSION
+        fi
+        git push origin $VERSION
```
nemesifier, line 48: What happens here if we ever need to push force on master?

openwisp_utils/__init__.py
```diff
@@ -1,4 +1,4 @@
-VERSION = (1, 1, 1, 'final')
+VERSION = (1, 2, 1, 'alpha')
```
nemesifier, line 1: why 1.2.1?

nemesifier (changes requested):
Is there a way to create a reusable github action so we don't need to duplicate this in each repo? That would allow us to change this for all repos from a single place.

pandafy:
> Is there a way to create a reusable github action so we don't need to duplicate this in each repo? That would allow us to change this for all repos from a single place.

We will need to find a way to automatically get the module name. Maybe, we can set this as an input parameter in GitHub workflow.

nemesifier (approved):

PR #495, opened 2025-07-25
[feature] Collect metric on opt-out #488

### LINKED ISSUE #488: [feature] Collect metric collection opt out action
When users opt out from collecting metrics from the web UI, it would be useful to get this last data point (only the opt out action and nothing else) before interrupting communications, so that we can figure out whether users are opting out or are stopping to use OpenWISP.

pandafy:
## Checklist

- [ ] I have manually tested the changes proposed in this pull request.

Closes #488

### DIFF
--- a/docs/user/metric-collection.rst
+++ b/docs/user/metric-collection.rst
@@ -36,7 +36,15 @@ Opting Out from Metric Collection
 ---------------------------------
 
 You can opt-out from sharing this data any time from the "System Info"
-page. Alternatively, you can also remove the
+page.
+
+.. note::
+
+    When a user opts out of metric collection using the web interface, a
+    one-time metric is still sent to record the opt-out event. This helps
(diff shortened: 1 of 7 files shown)


docs/user/metric-collection.rst
```diff
@@ -36,7 +36,15 @@ Opting Out from Metric Collection
...
 
 You can opt-out from sharing this data any time from the "System Info"
-page. Alternatively, you can also remove the
+page.
+
+.. note::
+
+    When a user opts out of metric collection using the web interface, a
+    one-time metric is still sent to record the opt-out event. This helps
+    us understand user preferences.
```
nemesifier, line 45: When a user opts out of metric collection through the web interface,
a one-time metric is still sent to record the opt-out event.
This helps distinguish between user abandonment and intentional opt-outs.

PR #497, opened 2025-07-28
[feature] Generate CHANGES.rst automatically

### LINKED ISSUE #496: [feature] Automate release process
Our current release process is entirely manual. It often involves a lot of time spent digging through git logs, manually writing and formatting the changelog, and remembering a sequence of git commands to tag and release a new version. This process can be time-consuming and creates friction that can discourage us from making frequent releases.

### The Proposal

To make this easier and faster, We can use a helper script to automate the most repetitive parts of making a release. The goal wouldn't be to remove humans from the loop. A manual review is still essential, but to handle the tedious, error-prone tasks for us.

### Proposed Core Functionality

@nemesifier and I put some thought into what a first version of this script could do. The main focus would be on perfecting the changelog generation. As discussed, it should be able to:

- Automatically scan git history since the last release tag.
- Understand our commit conventions, like [feature] etc.
- Handle breaking changes by looking for a ! (i.e., [change!]).
- Group commits into logical sections: Features, Changes (Backward incompatible changes, Other changes),  Dependencies and Bugfixes.
- Be smart about dependencies, showing only the latest version of a package if it was updated multiple times.
- Allow to skip commits from the changelog by using a flag like [skip changelog] on commit description.

Dhanus3133:
Closes #496


tests/test_generate_changelog.py
```diff
@@ -0,0 +1,103 @@
+import os
```
nemesifier, line 1: Please create a directory for the release tool, let's give this thing a name, eg: `releaser` or any other nice name you can think of. We'll concentrate all the test files for the different aspect of this tool in this directory.

tests/test_generate_changelog.py
```diff
@@ -0,0 +1,103 @@
...
+        self._git_commit("[test] Add tests for the new feature")
+        self._git_commit("[deps] Update django requirement from ~=3.2 to ~=4.0")
+
+        # Expected output
+        expected_unformatted_output = dedent(
+            """
+            Changelog
+            =========
+
+            [unreleased]
```
nemesifier, line 66: There's the version and date?

tests/test_generate_changelog.py
```diff
@@ -0,0 +1,103 @@
...
+            ~~~~~~~~~
+
+            - Correct a critical bug #123
+        """
+        ).strip()
+
+        processed_output = format_with_docstrfmt_file(
+            process_changelog(run_git_cliff())
+        )
+        formatted_expected = format_with_docstrfmt_file(expected_unformatted_output)
```
nemesifier, line 101: Wouldn't it better to preformat the change log and write it in the tests already well formatted?

tests/test_generate_changelog.py
```diff
@@ -0,0 +1,103 @@
...
+            Dependencies
+            ++++++++++++
+
+            - Bumped ``django~=4.1``
+            - Bumped ``requests<2.30``
+
+            Bugfixes
+            ~~~~~~~~~
+
+            - Correct a critical bug #123
```
nemesifier, line 94: Why don't we move these samples to rst files in a subdirectory, as we already do a lot in openwisp for sample test data? This has several advantages:

- test classes are shorter
- sample files are QA checked and formatted automatically
- we can use quite complex cases

We could do the same for the list of commits: provide text files for the list of commits, so we can easily create test samples with real data from our actual modules, everytime we find a bug or need to fix corner cases we can just copy/paste the commit log into a sample file and write a test for it in a short time.

nemesifier (commented):
Looks like it's slowly moving in the right direction.

See my comments below.

tests/test_generate_changelog.py
Dhanus3133: Updated!

releaser/tests/test_version_bumping.py
```diff
@@ -0,0 +1,36 @@
+from unittest.mock import mock_open, patch
+
+import pytest
+
+from releaser.version import bump_version, get_current_version
+
+SAMPLE_INIT_FILE = """
+VERSION = (1, 2, 0, "alpha", 0)
```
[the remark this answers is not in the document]
Dhanus3133: Done

releaser/github.py
```diff
@@ -0,0 +1,53 @@
...
+    def create_pr(self, head, base, title):
+        """Creates a pull request on GitHub."""
+        url = f"{self.base_url}/pulls"
+        payload = {
+            "title": title,
+            "head": head,
+            "base": base,
+            "body": "This PR was automatically generated by the release script.",
+        }
+        response = requests.post(url, headers=self.headers, json=payload)
```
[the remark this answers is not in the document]
Dhanus3133: Updated

nemesifier (commented):
I tried this now for this: https://github.com/openwisp/ansible-wireguard-openwisp/pull/71

The releaser.toml file got included in the commit:

Screenshot from 2025-08-22 16-53-41

Is this normal? I don't think this should happen.

I'll remove it manually for now.

Dhanus3133:
I don’t think `releaser.toml` is created automatically. From what I remember in the call, we were testing with the previous implementation (`releaser.toml` way), so I believe it was created at that time. Are you sure if there were no unstaged changes already present before running the release now?

PS: I think we could also add logic to notify the user about other changes that are available locally ater the checks.

nemesifier (changes requested):
@Dhanus3133 

> I don’t think `releaser.toml` is created automatically. From what I remember in the call, we were testing with the previous implementation (`releaser.toml` way), so I believe it was created at that time. Are you sure if there were no unstaged changes already present before running the release now?
> 
> PS: I think we could also add logic to notify the user about other changes that are available locally ater the checks.

I had to make some changes after the changes were created before proceeding: I had to reformat files due to the fact that we updated openwisp-utils to use `prettier` also for JSON and YAML files. So I pushed this change first, then I went ahead with the process after verifying the change log appeared.

Question: is it possible to avoid creating the file at all or is it really needed? In the worst case scenario we can add a .gitignore file in each repo although this will be time consuming. 

###  Ansible WLP testing error due to rst syntax issue

I tried this here: https://github.com/openwisp/ansible-openwisp-wifi-login-pages.

This is what happened:

```
Warning: Could not format content with `docstrfmt`.
WARNING: File "/tmp/tmpbw5uy424.rst", line 27:
Bullet list ends without a blank line; unexpected unindent.
Failed to format '/tmp/tmpbw5uy424.rst'
1 file was checked.
Done, but 1 error occurred ❌💥❌
```

Screenshot from 2025-08-22 17-32-23

This is probably caused by a restructured text syntax error somewhere which is crashing docstrfmt. However, it's not clear whether I can recover from this error manually. 

When this happens, it would be good to let the user know about the problem and return the error but we should also return the unformatted change log output, so the user can try figuring out where the problem is or in the worse case scenario overcome the issue by accepting the unformatted change log and then solving it manually before the final acceptance.

Dhanus3133:
> Question: is it possible to avoid creating the file at all or is it really needed? In the worst case scenario we can add a .gitignore file in each repo although this will be time consuming.

I am sure it was not created automatically. I tested out locally now. You can check this Dhanus3133/ansible-wireguard-openwisp#1 out.

Dhanus3133:
> ### Ansible WLP testing error due to rst syntax issue
> I tried this here: https://github.com/openwisp/ansible-openwisp-wifi-login-pages.
> 
> This is what happened:
> 
> ```
> Warning: Could not format content with `docstrfmt`.
> WARNING: File "/tmp/tmpbw5uy424.rst", line 27:
> Bullet list ends without a blank line; unexpected unindent.
> Failed to format '/tmp/tmpbw5uy424.rst'
> 1 file was checked.
> Done, but 1 error occurred ❌💥❌
> ```

Weird, I tried this on my fork now Dhanus3133/ansible-openwisp-wifi-login-pages#1 didn't seem to encounter this issue.
image


> When this happens, it would be good to let the user know about the problem and return the error but we should also return the unformatted change log output, so the user can try figuring out where the problem is or in the worse case scenario overcome the issue by accepting the unformatted change log and then solving it manually before the final acceptance.

Okay, understood.

PR #542, opened 2025-12-07
Prevent Duplicate Metric Usage Reports by Adding Daily Payload Hashing #531

### LINKED ISSUE #531: [feature] Avoid sending duplicate usage metrics
⚠️ **Not suitable to beginners**

Currently, the metric usage data may be sent multiple times per day if the Django project is configured incorrectly. i

I suggest that metric usage should be sent only once per day. To achieve this, we can calculate and store a hash of the payload along with the current date. On any given date, if the same payload is generated again, the hash should be checked before sending. If an identical hash (for the same date) already exists, skip sending to avoid duplication. This ensures each unique payload is sent at most once per day.

AditiSah05:
## Checklist

Closes #531 .

Please [open a new issue](https://github.com/openwisp/openwisp-utils/issues/new/choose) if there isn't an existing issue yet.

## Description of Changes
This pull request introduces a reliable mechanism to ensure that metric usage data is sent only once per day, addressing the issue of duplicate submissions caused by repeated executions in certain Django configurations.

To achieve this, the system now generates a hash of the daily metric payload and stores it along with the current date. When metrics are generated again on the same day, the stored hash is compared with the new one. If the payload is unchanged, the system simply skips the transmission. This prevents unnecessary duplication while keeping the reporting behavior consistent and lightweight.

In addition to the core logic, the following updates have been made:

-Added a persistent storage step for recording the daily payload hash.

-Implemented comparison logic to detect and prevent repeated submissions within the same day.

-Added and updated test cases to validate the new behavior, including hash creation, daily checks, and skip conditions.

-Updated the documentation to clearly explain the new flow and provide guidance on correct configuration.

## Screenshot
This change does not modify any user-facing interface, so no screenshot is required.

### DIFF
--- a/.github/workflows/reusable-version-branch.yml
+++ b/.github/workflows/reusable-version-branch.yml
@@ -16,19 +16,26 @@ on:
 jobs:
   replicate:
     runs-on: ubuntu-latest
+    permissions:
+      # required for check-out and push
+      contents: write
+      # required for getting an authentication token
+      id-token: write
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v6
+        # Use actions/checkout@v4, the current major version
+        uses: actions/checkout@v4
+        with:
+          # Fetch all history so rebase can work without a separate fetch step
+          fetch-depth: 0
 
       - name: Set up Python
-        uses: actions/setup-python@v6
+        uses: actions/setup-python@v5
         with:
           python-version: "3.10"
-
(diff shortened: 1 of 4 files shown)


AditiSah05:
Hi @pandafy ,
I’ve completed the changes and submitted the pull request. Whenever you have a moment, could you please review it and share your feedback?
Thank you!

PR #553, opened 2025-12-31
[releaser] Fix git add to only include release files #552

### LINKED ISSUE #552: [bug] Releaser tool adds any change and pushes it
**Describe the bug**

The releaser tool adds all the files for being committed instead of only adding the files which have been changed.
This is wrong.

I think the problem stems from:

https://github.com/openwisp/openwisp-utils/blob/master/openwisp_utils/releaser/release.py#L361-L362

This dangerous as files which are meant to be private may be pushed by mistake to the repository.

It's ok to add files which have been changed by the developer (not only by the releaser tool itself) as devs may do minor changes to different files while preparing the release.

Srinath0916:
The releaser tool was using 'git add .' which dangerously adds all files including potentially private ones. Changed to only add specific files that are modified during the release process (changelog and version files).

Fixes #552

## Checklist

Closes #552.

## Description of Changes

I noticed the releaser tool had a security issue where it was using `git add .` to stage files for commit. This is problematic because it adds every single file in the repository, including things like private config files or temporary files that shouldn't be committed.

The fix was straightforward - instead of adding everything, I made it only add the specific files that actually get modified during a release:
- The changelog file (which always gets updated)
- The version file (if it exists and gets modified)

I also added some comments to explain why we're being selective about which files to add, and updated the console message to be clearer about what's happening.

The change is in the `openwisp_utils/releaser/release.py` file around lines 361-370. I tested it by running the code quality checks and making sure the syntax is correct.

This should prevent any accidental commits of private files during the release process.

## Screenshot
Not applicable since this is a backend code fix.

## Summary by CodeRabbit

* **Chores**
  * Release flow now stages only modified, tracked files instead of all working-tree changes, producing more precise release commits while leaving commit/push/PR steps and overall process unchanged.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/openwisp_utils/releaser/release.py
+++ b/openwisp_utils/releaser/release.py
@@ -358,17 +358,8 @@ def main():
         ["git", "checkout", "-b", release_branch], check=True, capture_output=True
     )
 
-    print("Adding release-related files to git...")
-    # Add only the specific files that were modified during the release process
-    files_to_add = [config["changelog_path"]]
-
-    # Add version file if it exists and was modified
-    version_path = config.get("version_path")
-    if version_path and os.path.exists(version_path):
-        files_to_add.append(version_path)
-
-    for file_path in files_to_add:
-        subprocess.run(["git", "add", file_path], check=True, capture_output=True)
+    print("Adding tracked changes to git...")
+    subprocess.run(["git", "add", "-u"], check=True, capture_output=True)
 
     commit_message = f"{new_version} release"
     subprocess.run(


Srinath0916:
Hi @nemesifier , I've implemented a fix for the security issue you reported. The dangerous `git add .` has been replaced with targeted file additions to only include changelog and version files during the release process. Ready for review when you have time. Thanks!

nemesifier (commented):
Looks good, I'll need to test it on a new bugfix release before merging.

Srinath0916:
Heyy @nemesifier , i have updated my PR as you suggested, please have a look.

Issue #523, opened 2025-10-28
[feature] Have AI propose change log entry / commit message


nemesifier:
Follow up of the efforts to make the release process easier and free up the time of maintainers (https://github.com/openwisp/openwisp-utils/issues/496), related to https://github.com/openwisp/openwisp-utils/issues/501, https://github.com/openwisp/openwisp-utils/issues/502.

When a PR is flagged as approved by a maintainer and submitted with a specific comment (eg: we could mention a bot or write a specific tag), we could get an AI bot or anything along these lines to analyze the PR, the implementation and suggest a change log entry in RestructuredText format in a github comment (formatted as code) which can be copied/pasted by the maintainer in the github web interface (commit message long description) when squash merging.

The long description will be picked up later on by the releaser tool (#496) and used for the change log.

There may be bots who can do this sort of stuff already, help is welcome to research this topic!

pushpitkamboj:
I am working on this, will soon open a PR on it!

PR #560, opened 2026-01-01
[feature/qa] Standardize commit messages using commitizen #110

### LINKED ISSUE #110: [qa] Enforce conventional commit style

Stoiicc:
## Checklist

Closes #110.

## Description of Changes

This PR introduces a standardized commit message workflow for OpenWISP by integrating Commitizen into openwisp-utils. The goal is to make commit messages more consistent, easier to review, and simpler for contributors to write correctly. This change adds a guided, interactive commit workflow using the `cz commit` command. Contributors are prompted step by step to generate a commit message that follows OpenWISP conventions.

The workflow enforces:
- a consistent commit message structure
- a properly formatted commit title
- inclusion of an issue reference in the title
- automatic generation of a Fixes #<issue_number> footer

Allowed commit prefixes:

  - feature
  - change
  - fix
  - docs
  - test
  - ci
  - chores
  - qa
  - deps
  - release
  - bump

If none of the predefined prefixes apply, contributors can select "chores".

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ setup(
             "coverage>=7.10.0,<7.14.0",
             "tblib~=3.2.2",
             "docstrfmt>=2.0.0,<2.1.0",
-            "commitizen>=4.0.0,<5.0.0",
+            "commitizen>=4.13.0,<5.0.0",
         ],
         "rest": [
             "djangorestframework~=3.16.0",


nemesifier (commented):
This is definitely interesting! We're being flooded with PRs but I am doing my best to review most of them. I will come back to this asap, in the meantime see my comments below.

nemesifier:
See also https://github.com/openwisp/openwisp-utils/issues/523, if we get the work in this PR right we'll definitely use it as a base for that issue.

nemesifier (commented):
Great work @Stoiicc, I am really glad you took on this issue that has been opened for a long time and I neglected it as well, you motivivated me to invest a bit of time in this and I hope it will help newcomers passing QA checks faster.

I did some changes to make sure all cases are covered, there may be some tweaks needed down the road but we can always open new PRs to fix any issue we find, so it looks good to me now, but **please double check my changes**, let me know if you have any doubt/question.

nemesifier (dismissed):

nemesifier:
@Stoiicc thanks, addressed in https://github.com/openwisp/openwisp-utils/pull/560/commits/6f0fc27a3c930b1d392f26d99c523f7388e8a29c.

PR #596, opened 2026-02-24
[change] Allow cascade deletions in ReadOnlyAdmin

### LINKED ISSUE #256: [bug] Deleting organization is prevented by Mass Upgrade Operation
- Create an org
- Create a dummy firmware image in the new org
- Launch a mass upgrade operation for the firmware image (alternatively use the django shell to create it) and org created in the previous steps
- Now try to delete the org

**Expected outcome:** 

- The confirmation page is shown
- Once the operation is confirmed, the delete is carried out successfully

**Actual outcome:**

> Cannot delete organization
> Deleting the selected organization would result in deleting related objects, but your account doesn't have permission to delete the following types of objects:
>
> - Mass upgrade operation

Screenshot from 2024-05-15 17-29-12

UltraBot05:
## Checklist

Closes [#256](https://github.com/openwisp/openwisp-firmware-upgrader/issues/256)

## Description of Changes

`ReadOnlyAdmin.has_delete_permission` previously returned `False` unconditionally, which blocked Django's cascade delete machinery when a parent model (e.g. Organization) tried to delete a child read-only model.

This PR introduces `BlockDeleteAllowCascadeMixin` and composes `ReadOnlyAdmin` from it, making the mixin reusable by other modules.

**Key changes:**
- New `BlockDeleteAllowCascadeMixin`: blocks deletions on the model's own `_delete`, `_change`, and `_changelist` views; allows cascade-triggered checks from parent delete views and `delete_selected` bulk actions.
- `getattr` guards on `resolver_match` prevent crashes in test/non-request contexts.
- `ReadOnlyAdmin` now inherits from the mixin instead of defining `has_delete_permission` directly.
- Tests cover direct admin routes, cascade contexts, `None` resolver, staff-without-delete-permission, and a full integration test deleting a parent through `ReadOnlyAdmin` children end-to-end.

Companion PR applying this to [firmware-upgrader#391](https://github.com/openwisp/openwisp-firmware-upgrader/pull/391)

## Screenshot
Screenshot 1 Shows the deletion of org -> Failure page
Screenshot 2 Shows the working Fix

#256-before
#256-after


nemesifier (commented):
@UltraBot05 see https://github.com/openwisp/openwisp-firmware-upgrader/issues/256#issuecomment-3953193076

nemesifier (approved):
Approving to trigger and test the changelog bot, this PR seems the most ready of the currently open one so I am testing here, not confirmed ready to merge yet. 

CC: @pushpitkamboj

nemesifier:
I pushed one more commit.

Changes:

- added a concise comment in `ReadOnlyAdmin.has_delete_permission` to explain why the URL checks are there;
- made the logic stricter: direct delete/change/changelist of the read-only model is still blocked, and we only fall back to Django's normal delete permission check when the call comes from another admin delete view or another changelist `delete_selected` action;
- added an integration test which registers a cascaded child model with `ReadOnlyAdmin` and verifies deleting the parent from admin works end to end;
- restored `get_user_model()` in tests.

Next steps:

- apply the same stricter approach in openwisp-firmware-upgrader#391, preferably by sharing the logic or keeping the behavior aligned;
- test again in a full OpenWISP environment. If org deletion still reports `Device`, that is a separate blocker in the device admin and should be fixed in the related app.

UltraBot05:
Okay, Resolved the merge conflict with upstream/master - it was a small _import-section_ conflict in `tests/test_project/tests/test_admin.py` (lines 3–7, `django.contrib.admin` vs `django.contrib.admin.SimpleListFilter`), kept both imports. Branch is now up to date with upstream. Ready for review.

Edit: Also updated the stale PR Description

nemesifier (approved):

PR #614, opened 2026-03-11
Fix email template dark mode rendering by forcing light theme

### LINKED ISSUE #490: [bug] Email template does not support dark theme
**Describe the bug**
Some email clients force the dark theme on the email content. Due to this, the logo becomes nearly invisibly in dark theme 

**Expected outcome**
We need to find a way to override the dark theme in this scenario and force light theme. 

Image

aswinms926:
## Checklist

Closes #490

## Description of Changes

This change fixes the issue where some email clients automatically force dark mode
on email content, causing the logo to become nearly invisible.

The email template is updated to explicitly declare a light color scheme so that
email clients do not apply automatic dark mode transformations.

This ensures the email background remains white and the logo stays visible even
when the email client is set to dark mode.


nemesifier (commented):
Thanks @aswinms926, this is a sensible fix. Forcing a light theme in transactional emails is the right call because we need consistent rendering across clients more than we need to honor dark mode.

I checked the diff and the approach looks correct, but I spotted one small issue:

- There is trailing whitespace on the blank line right before the `[data-ogsc]` rule. Please clean it up.

Also, please confirm that you have manually tested the email in the major clients (Gmail web/app dark mode, Apple Mail dark, Outlook) and that the logo and text stay legible. Since this is hard to cover with automated tests, a quick manual check and a note in the PR is enough.

With the whitespace cleaned and the manual check confirmed, this is good to merge.

pandafy (commented):
I have tested this fix on Gmail android app and web app. It forces the template to use the light theme as expected.

PR #570, opened 2026-01-29
[change] Handle skipped tests in SeleniumTestMixin

asmodehn:
## Checklist

- [X] I have read the [OpenWISP Contributing Guidelines](http://openwisp.io/docs/developer/contributing.html).
- [X] I have manually tested the changes proposed in this pull request.
- [ ] I have written new test cases for new code and/or updated existing tests for changes to existing code.
- [ ] I have updated the documentation.

## Reference to Existing Issue

Closes #569 .

## Description of Changes

Here we check if the test has been skipped, to correctly report it as skipped.
It s a quick patch that worked on my setup. Feel free to adjust.

## Related concerns: 
We might want to double check the behavior of an `@expectedFailure` test marker...

PR #620, opened 2026-03-12
[fix] Handle skipped Selenium tests in Django parallel runner

### LINKED ISSUE #619: [bug] SeleniumTestMixin skip handling crashes under Django parallel runner
## Summary

`SeleniumTestMixin._setup_and_call()` crashes under Django's parallel test runner when a Selenium test is skipped.

The current skip handling added in `openwisp-utils#570` iterates `result.skipped`, but Django's `RemoteTestResult` sets `skipped` to a truthy `DummyList`, which is not iterable.

## Steps to reproduce

1. Run a skipped Selenium test through Django's parallel runner.
2. Let the test reach `SeleniumTestMixin._setup_and_call()`.

## Current behavior

The test run crashes in `openwisp_utils/tests/selenium.py` with:

```python
TypeError: 'DummyList' object is not iterable
```

at:

```python
for _, reason in result.skipped:
```

## Expected behavior

Skipped Selenium tests should still be reported as skipped under Django's parallel runner, without crashing the test run.

## Root cause

For Django's parallel runner, the result object is `RemoteTestResult`:

- `result.skipped` is a `DummyList`
- `DummyList` is truthy but not iterable
- actual skip events are recorded in `result.events`

So the current logic works for standard unittest results, but not for `RemoteTestResult`.

## Proposed fix

- read skip reasons from `result.events` when the result object exposes `events`
- fall back to iterating `result.skipped` for standard results

This preserves the behavior added in `#570` for serial runs and fixes the parallel case.

czarflix:
## Checklist

- [X] I have read the [OpenWISP Contributing Guidelines](http://openwisp.io/docs/developer/contributing.html).
- [X] I have manually tested the changes proposed in this pull request.
- [X] I have written new test cases for new code and/or updated existing tests for changes to existing code.

Closes #619.

## Description of Changes

`openwisp-utils#570` fixed skipped Selenium tests by propagating skips from `result.skipped`.

That works for standard unittest results, but it breaks under Django's parallel test runner because `RemoteTestResult.skipped` is a truthy `DummyList`, which is not iterable. In that case, skip information is recorded in `result.events`.

This patch keeps the serial skip behavior introduced by `#570` and fixes the parallel case by reading skip reasons from `result.events` when available, while falling back to `result.skipped` for standard results.

It also adds focused regression coverage for both result types:

- standard `TestResult`
- Django `RemoteTestResult`

## Verification

I verified locally with:

- `./run-qa-checks`
- `python tests/manage.py test test_project.tests.test_selenium_mixin`
- focused parallel runs of `test_project.tests.test_selenium_mixin` under Django `4.2`, `5.0`, `5.1`, and `5.2`

I also checked the behavior around `#570` directly:

- before `#570`, skipped Selenium tests were incorrectly reported as successful
- `#570` fixed serial skip propagation but crashes under the parallel result type
- this patch preserves the serial fix and handles the parallel result correctly

## Screenshot

N/A

### DIFF
--- a/tests/test_project/tests/test_selenium_skip.py
+++ b/tests/test_project/tests/test_selenium_skip.py
@@ -51,13 +51,11 @@ class TestSeleniumMixinSkipHandling(SimpleTestCase):
 
     def test_setup_and_call_propagates_skip_to_standard_result(self):
         result = self._run_skipped_standard_test()
-
         self.assertEqual(len(result.skipped), 1)
         self.assertEqual(result.skipped[0][1], "skip propagation regression test")
 
     def test_setup_and_call_preserves_remote_skip_events_for_multiple_tests(self):
         result = self._run_skipped_remote_suite()
-
         self.assertEqual(
             result.events,
             [


czarflix:
@asmodehn Should I push the changes ?

czarflix:
hi @nemesifier 

> @coderabbitai review

All Green

czarflix:
Follow-up pushed. The event-log corruption concern was correct: the skip path was overwriting the shared `RemoteTestResult` event log.

I removed that overwrite, record `stopTest()` on the outer remote result instead, and updated the regression coverage to run multiple skipped methods from the same `SeleniumTestMixin` class through `RemoteTestRunner`, so the shared-worker case is now covered.

nemesifier (commented):
Code looks good but manual testing is needed before confirming and merging.

czarflix:
> Looks like a good and clean set of changes with matching test. Thanks !

@nemesifier , please take a look when you have the time

czarflix:
@nemesifier  
For manual verification, I tested the skip path locally in both serial and Django parallel runs, including the shared RemoteTestResult case with multiple skipped methods in the same class. The behavior is now correct in both cases and the shared event log is preserved. If you want, I can rerun any specific scenario.

czarflix:
> I did some minor changes, the most important one being to always call `original_result.stopTest(self)` when the test has to be skipped, while before this call was conditional, which didn't make sense: if it's skipped it should be always be stopped, right? Let me know if that's not the case @czarflix @asmodehn.

I think this change is correct.

The key point is that `_setup_and_call()` manually calls `original_result.startTest(self)` before the retry loop, and neither `SimpleTestCase.__call__()` nor `TestSuite.run()` will balance that on the outer result for us later. In the normal unittest contract, a skipped test still goes through `startTest -> addSkip -> stopTest`, and for `RemoteTestResult` the `stopTest()` call is also what records the final `("stopTest", index)` event.

So I don’t see a case where a propagated skip on `original_result` should skip `stopTest()`.

PR #629, opened 2026-03-17
[chores] Cache CI bot dependencies #628

### LINKED ISSUE #628: [chores] CI bots: cache dependencies
The bots always reinstall all the dependencies they need at each run, a huge waste of resources (hardware, network bandwidth, energy, time). We need to cache the dependencies, for the pip deps we can use `setup.py` for the hash so if anything changes there the cache is invalidated.

czarflix:
## Checklist

Closes #628.

## Description of Changes

- enabled pip caching for the bot workflows that install `openwisp-utils[github_actions]`, using `setup.py` as the cache invalidation key
- added cache support to the changelog composite action without changing its install command, and passed a workspace-relative dependency path from the reusable workflow
- updated the changelog workflow sparse checkout to include `setup.py` so the cache key is available there as well

Local validation:

- `git diff --check`
- `actionlint .github/workflows/*.yml`
- `act` validation covering the repo-root, nested checkout, `trusted_scripts`, sparse checkout, and composite-action caller-path shapes

## Screenshot

N/A, workflow-only change.

### DIFF
--- a/.github/actions/bot-changelog-generator/action.yml
+++ b/.github/actions/bot-changelog-generator/action.yml
@@ -15,11 +15,24 @@ inputs:
   repo-name:
     description: "Repository name in format owner/repo"
     required: true
+  cache-dependency-path:
+    description: "Path to dependency metadata used for pip cache invalidation"
+    required: false
+    default: ""
 
 runs:
   using: "composite"
   steps:
+    - name: Set up Python with cache
+      if: ${{ inputs.cache-dependency-path != '' }}
+      uses: actions/setup-python@v5
+      with:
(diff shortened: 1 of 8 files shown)


nemesifier (commented):
The goal is to cache the installation of openwisp-utils itself, so we shouldn't need to add a new input parameter as the result is always the same in all repos. Are you sure this is the correct approach?

The problem is that now we're installing openwisp-utils directly and using master, we may need to do something different to make cache invalidation easier.

pandafy:
@czarflix @nemesifier what is the status on this?

PR #642, opened 2026-04-07
[feat] Add Select2Widget for choice fields #254

### LINKED ISSUE #254: [feature] Add reusable solution for turning choice fields into select2 autocomplete fields
We should look for a way to turn this into a reusable solution: https://github.com/openwisp/openwisp-firmware-upgrader/pull/158.

It is possible that there are already some solutions out there, we should investigate.

shivsubh:
Implement a reusable Select2Widget that uses Django's native admin assets to avoid extra dependencies.

Closes #254

## Checklist

Closes #254 .

### DIFF
--- a/tests/test_project/admin.py
+++ b/tests/test_project/admin.py
@@ -7,6 +7,7 @@ from openwisp_utils.admin import (
     HelpTextStackedInline,
     ReadOnlyAdmin,
     ReceiveUrlAdmin,
+    Select2AdminMixin,
     TimeReadonlyAdminMixin,
     UUIDAdmin,
 )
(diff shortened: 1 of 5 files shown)


tests/test_project/admin.py
```diff
@@ -122,6 +123,11 @@ class ShelfAdmin(TimeReadonlyAdminMixin, admin.ModelAdmin):
     ]
     search_fields = ["name"]
 
+    def formfield_for_choice_field(self, db_field, request, **kwargs):
+        if db_field.name == "books_type":
+            kwargs["widget"] = Select2Widget()
+        return super().formfield_for_choice_field(db_field, request, **kwargs)
```
pandafy: We want a re-usable solution, and if we go by this implementation then we will have to override the `formfield_for_choice_field` method everytime we want to have a ModelAdmin field use a select2 widget. 

Can we think about adding a class variable, similar to `readonly_fields` where we can define which fields should use this widget. Moreover, we need to ensure that this widget can only be used on choice fields, otherwise it may break the form.

pandafy (changes requested):
Thank you @shivsubh for contributing. We want to have a re-usable solution which means minimal code changes should be required in the downstream projects to utilize the new widget. 

I did a quick review and found the following code problematic. 

We will need new selenium tests which verifies the functioning of the added field.

tests/test_project/admin.py
shivsubh: Thanks for the review! @pandafy. I have addressed the approach that you advised in the last commit. You can review it and tell me if it is correct or need some refactoring.

PR #650, opened 2026-04-19
[ci] Fixed CI transient error re-run mechanism

stktyagi:
## Checklist

## Description of Changes

The auto-retry failed to trigger on transient infrastructure crashes (like network drops) because their log outputs contained the word "ERROR:". Since "ERROR:" was globally listed as a test failure marker, the script falsely flagged these transient issues as real code bugs and aborted the retry.  (e.g., `"ERROR: Could not install packages due to an OSError"`).
To fix this:
Failure markers are categorized into Strict (e.g., FAIL:, which always block retries) and Generic (e.g., "ERROR:", which can be bugs or infrastructure crashes). Now, if the script detects a known transient crash, it safely "forgives" any generic markers and successfully triggers the auto-retry.

### DIFF
--- a/.github/actions/bot-ci-failure/analyze_failure.py
+++ b/.github/actions/bot-ci-failure/analyze_failure.py
@@ -149,10 +149,15 @@ def process_error_logs(content):
             continue
         seen_bodies.add(body_key)
         total_unique_jobs += 1
-        job_has_test_failure = any(m in body for m in TEST_FAILURE_MARKERS)
-        if _is_transient_failure(body) and not job_has_test_failure:
+        is_transient = _is_transient_failure(body)
+        cleaned_body = body
+        if is_transient:
+            for t_marker in TRANSIENT_FAILURE_MARKERS:
+                pattern = re.compile(re.escape(t_marker), re.IGNORECASE)
+                cleaned_body = pattern.sub("", cleaned_body)
+        job_has_test_failure = any(m in cleaned_body for m in TEST_FAILURE_MARKERS)
+        if is_transient and not job_has_test_failure:
             transient_jobs += 1
-        # Detect real test failures and keep only the failing parts.
         if job_has_test_failure:
             tests_failed = True
             body = _extract_failed_tests(body)
@@ -368,6 +373,9 @@ def main():
          issue" and mention the CI has been restarted automatically if applicable.
        - For Coveralls failures specifically, mention it is a known flaky
          service and not actionable by the contributor.
+       - If there are ALSO real test failures (like AssertionErrors) in the logs,
+         tell the contributor to fix the real test failures first and push a new commit.
+         Do NOT tell them the CI restarted automatically if real test failures exist.
 
     5. **Build/Infrastructure/Other** (missing dependencies, Docker errors,
        setup failures that are NOT transient)
--- a/.github/actions/bot-ci-failure/analyze_failure.py
+++ b/.github/actions/bot-ci-failure/analyze_failure.py
@@ -6,15 +6,22 @@ import sys
 from google import genai
 from google.genai import types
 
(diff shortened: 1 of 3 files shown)


stktyagi:
@coderabbitai review

stktyagi:
tested retry mechanism [here](https://github.com/openwisp/bot-testing-ground/pull/40#issuecomment-4276578664)


image

------------------------------------------------------------------------------------------
image

.github/actions/bot-ci-failure/analyze_failure.py
```diff
@@ -149,10 +149,15 @@ def process_error_logs(content):
             continue
         seen_bodies.add(body_key)
         total_unique_jobs += 1
-        job_has_test_failure = any(m in body for m in TEST_FAILURE_MARKERS)
-        if _is_transient_failure(body) and not job_has_test_failure:
+        is_transient = _is_transient_failure(body)
```
nemesifier, line 159: There's a bunch of lines of code here and zero explanation of why this is needed. We need a preceding comment which summarizes why this is needed.

.github/actions/bot-ci-failure/analyze_failure.py
```diff
@@ -149,10 +149,15 @@ def process_error_logs(content):
...
+        cleaned_body = body
+        if is_transient:
+            for t_marker in TRANSIENT_FAILURE_MARKERS:
+                pattern = re.compile(re.escape(t_marker), re.IGNORECASE)
+                cleaned_body = pattern.sub("", cleaned_body)
```
nemesifier: Are you stripping out the transient marker right?

My concern is: the transient markers do not contain full errors, just part of it, does it make sense to strip something only partially?

nemesifier (commented):
@stktyagi this PR fails to explain the problem that is causing the issue in the first place.
Whenever you open a PR, please explain the problem carefully.

As I do not know what exact problem was causing the auto-restart to not work, I cannot confidently know if this solves it properly.

stktyagi:
> @stktyagi this PR fails to explain the problem that is causing the issue in the first place. Whenever you open a PR, please explain the problem carefully.
> 
> As I do not know what exact problem was causing the auto-restart to not work, I cannot confidently know if this solves it properly.

Understood! As there was no preceding issue for this I must have missed issue explanation properly. I'll update the the description and make sure going forward

.github/actions/bot-ci-failure/analyze_failure.py
stktyagi: I'll add comments above the new changes as explanation

.github/actions/bot-ci-failure/analyze_failure.py
stktyagi: As it was mostly an issue related to the test failure keywords blocking the transient re-run it does fix most failure cases but I asked AI to create edge cases where it would fail and one particular edge case is
A network crash prints: `"Traceback (most recent call last): ...`

Because of errors like this, instead of stripping we can try this approach -

1. We can categorize failure markers into "Strict" (FAIL:, AssertionError) which always blocks retries, and "Generic" (ERROR:, Traceback) which can result from either real bugs or infrastructure crashes.
2. Now, if a transient crash is detected, the bot would leave the logs completely unaltered and safely "forgives" the generic markers so the auto-retry can successfully trigger.

I'll test this once in bot-testing-ground and update the PR.

stktyagi:
@nemesifier 
I've manually tested for multiple failure cases [here](https://github.com/openwisp/bot-testing-ground/pull/96) in a new PR for readability.

stktyagi:
@coderabbitai review

nemesifier (approved):
Thanks @stktyagi 🙏

PR #652, opened 2026-04-22
[chores:fix] Fixed auto-assign bot claiming success on silent rejection

### LINKED ISSUE #649: [bug] Auto-assign bot claims assignment succeeded when it silently fails for non-collaborators
**Describe the bug**

When a non-organization-member opens a PR that references an issue, the auto-assign bot ([`issue_assignment_bot.py`](https://github.com/openwisp/openwisp-utils/blob/master/.github/actions/bot-autoassign/issue_assignment_bot.py)) calls `issue.add_to_assignees(pr_author)` and then immediately posts:

> This issue has been automatically assigned to @user who opened PR #N to address it. 🎯

However, GitHub has a strict security rule: an issue cannot be assigned to a user who is not a member of the organization, unless they have previously commented on that specific issue. When this restriction kicks in, the GitHub API silently ignores the assignment request (no exception is raised), but the bot still posts the comment claiming the assignment was successful.

**Example**

See [#646 (comment)](https://github.com/openwisp/openwisp-utils/issues/646#issuecomment-4225383930) — the bot commented that the issue was assigned to `@prathmeshkulkarni-coder`, but the assignment never actually happened.

**Proposed solution**

1. After calling `issue.add_to_assignees(pr_author)`, verify the assignment actually took effect (re-fetch assignees).

2. **If the assignment succeeded** → post the current confirmation message as-is.

3. **If the assignment failed silently** → post an honest fallback message, for example:

   > Hi @{pr_author} 👋,
   >
   > You have opened PR #{pr_number} to address this issue, but GitHub does not allow automatic issue assignment for users who are not yet organization members and have not previously commented on the issue.
   >
   > Once you comment on this issue, GitHub will allow the assignment. You can comment `@openwisp-companion assign` to have the bot assign you automatically.

4. Extend `handle_issue_comment` to recognize the `@openwisp-companion assign` command:
   - Check if the commenter has an open PR referencing this issue
   - If yes → call `add_to_assignees` (which will now succeed since they have commented) and post the confirmation message
   - If no → reply that no open PR referencing this issue was found

This way the bot comment accurately reflects what happened, and the contributor can self-serve the assignment without needing a maintainer to step in — while the act of mentioning the bot naturally satisfies the GitHub security rule.

**References**

- Discussion: https://github.com/openwisp/openwisp-utils/issues/646#issuecomment-4225383930
- Relevant code: [`auto_assign_issues_from_pr()`](https://github.com/openwisp/openwisp-utils/blob/master/.github/actions/bot-autoassign/issue_assignment_bot.py#L197-L252) and [`handle_issue_comment()`](https://github.com/openwisp/openwisp-utils/blob/master/.github/actions/bot-autoassign/issue_assignment_bot.py#L269-L291)

Eeshu-Yadav:
## Checklist

Closes #649

Verify assignment actually took effect before posting the confirmation comment. On silent rejection (non-collaborators who have not commented on the issue), post an honest fallback message instructing the user to trigger assignment via the "@openwisp-companion assign" bot command.

## Testing on `bot-testing-ground`

  Workflows on `openwisp/bot-testing-ground@master` temporarily point at this branch
  (commit [`7b24383`](https://github.com/openwisp/bot-testing-ground/commit/7b24383))
  so the silent-rejection and recovery paths can be exercised end-to-end by a
  non-org contributor.

  | Scenario | Issue | PR |
  |---|---|---|
  | Silent-rejection fallback + `@openwisp-companion assign` recovery | openwisp/bot-testing-ground#97 | openwisp/bot-testing-ground#98 |
  | Silent-rejection fallback + `@openwisp-companion assign` recovery (second run) | openwisp/bot-testing-ground#99 | openwisp/bot-testing-ground#100 |

### DIFF
--- a/.github/actions/bot-autoassign/issue_assignment_bot.py
+++ b/.github/actions/bot-autoassign/issue_assignment_bot.py
@@ -3,12 +3,12 @@ import re
 import time
 
 from base import GitHubBot
-from github import GithubException
 from utils import (
     extract_linked_issues,
     find_open_pr_for_issue,
     get_valid_linked_issues,
     unassign_linked_issues_helper,
+    user_in_logins,
 )
 
 VERIFICATION_RETRY_DELAY_SECONDS = 1.0
@@ -212,42 +212,30 @@ class IssueAssignmentBot(GitHubBot):
 
     @staticmethod
     def _get_assignee_logins(issue):
-        return [
-            assignee.login
-            for assignee in getattr(issue, "assignees", []) or []
-            if hasattr(assignee, "login")
-        ]
+        return [a.login for a in issue.assignees if hasattr(a, "login")]
 
-    @staticmethod
-    def _user_in(user, logins):
-        user_lower = (user or "").lower()
-        return any(user_lower == (login or "").lower() for login in logins)
-
-    def _assignment_succeeded(self, issue_number, user):
-        # GitHub silently rejects assignment of non-collaborators who
-        # have not commented on the issue. Re-fetch to confirm, and
-        # retry once to absorb read-after-write lag. Returns None when
-        # verification itself errored — caller must stay silent.
-        attempts = 2
-        last_error = None
-        for attempt in range(attempts):
+    def _verify_assignment(self, issue_number, user):
+        """Re-fetch to confirm `user` was assigned. Returns True/False
+        on verified state, None when every fetch errored — caller
+        stays silent on None since the true state is unknown.
+        """
+        had_successful_fetch = False
+        for attempt in range(2):
             try:
                 updated_issue = self.repo.get_issue(issue_number)
-                if self._user_in(user, self._get_assignee_logins(updated_issue)):
+                had_successful_fetch = True
+                if user_in_logins(user, self._get_assignee_logins(updated_issue)):
                     return True
-            except GithubException as e:
-                last_error = e
+            except Exception as e:
                 print(
-                    f"GitHub API error verifying assignment for #{issue_number}"
-                    f" (attempt {attempt + 1}/{attempts}): {e}"
+                    f"Error verifying assignment for #{issue_number}"
+                    f" (attempt {attempt + 1}/2): {e}"
                 )
-            if attempt + 1 < attempts:
+            if attempt == 0:
                 time.sleep(VERIFICATION_RETRY_DELAY_SECONDS)
-        if last_error is not None:
-            return None
-        return False
+        return False if had_successful_fetch else None
 
-    def _build_silent_failure_message(self, pr_author, pr_number):
+    def _cannot_auto_assign_message(self, pr_author, pr_number):
         return (
             f"Hi @{pr_author} 👋,\n\n"
             f"Thank you for opening PR #{pr_number} to address this issue!\n\n"
@@ -267,8 +255,7 @@ class IssueAssignmentBot(GitHubBot):
             issue = self.repo.get_issue(issue_number)
             if (
                 hasattr(issue, "repository")
-                and getattr(issue.repository, "full_name", self.repository_name)
-                != self.repository_name
+                and issue.repository.full_name != self.repository_name
             ):
                 print(
                     f"Issue #{issue_number} is from a different"
@@ -281,21 +268,29 @@ class IssueAssignmentBot(GitHubBot):
             if getattr(issue, "state", "open") == "closed":
                 print(f"#{issue_number} is closed, ignoring bot command")
                 return True
-            if self._user_in(commenter, self._get_assignee_logins(issue)):
+            if user_in_logins(commenter, self._get_assignee_logins(issue)):
                 print(f"{commenter} is already assigned to #{issue_number}")
                 return True
-            pr = find_open_pr_for_issue(self.repo, commenter, issue_number)
+            try:
+                pr = find_open_pr_for_issue(
+                    self.github, self.repository_name, commenter, issue_number
+                )
+            except Exception as e:
+                # Don't post "no PR found" on a search error — that's
+                # not the same as a verified miss.
+                print(f"Error searching open PRs by {commenter}: {e}")
+                return True
             if pr is None:
                 issue.create_comment(
                     f"Hi @{commenter} 👋,\n\n"
                     "I could not find an open PR by you that references"
                     f" this issue (#{issue_number}). Please open a PR"
                     f" linking to this issue (e.g. `Fixes #{issue_number}`)"
-                    " and then re-run the command."
+                    f" and then comment `@{self.bot_username} assign` again."
                 )
                 return True
             issue.add_to_assignees(commenter)
-            verified = self._assignment_succeeded(issue_number, commenter)
+            verified = self._verify_assignment(issue_number, commenter)
             if verified is True:
                 issue.create_comment(
                     f"This issue has been assigned to @{commenter}"
@@ -303,6 +298,8 @@ class IssueAssignmentBot(GitHubBot):
                 )
                 print(f"Assigned #{issue_number} to {commenter} via bot command")
             elif verified is False:
+                # Commenter has commented but the assignment still
+                # failed (perm block, outage, etc.).
                 issue.create_comment(
                     f"Sorry @{commenter}, GitHub still did not allow"
                     " this assignment. A maintainer will need to assign"
@@ -310,12 +307,12 @@ class IssueAssignmentBot(GitHubBot):
                 )
                 print(
                     f"Bot-command assignment of #{issue_number} to"
-                    f" {commenter} was silently rejected"
+                    f" {commenter} was silently rejected."
                 )
             else:
                 print(
                     f"Skipping comment for #{issue_number}:"
-                    " assignment state could not be verified"
+                    " assignment state could not be verified."
                 )
             return True
         except Exception as e:
@@ -339,7 +336,7 @@ class IssueAssignmentBot(GitHubBot):
                 )
             assigned_issues = []
             for issue_number, issue in get_valid_linked_issues(
-                self.repo, self.repository_name, pr_body
+                self.repo, self.repository_name, linked_issues
             ):
                 if len(assigned_issues) >= max_issues:
                     break
@@ -352,7 +349,7 @@ class IssueAssignmentBot(GitHubBot):
                         continue
                     current_assignees = self._get_assignee_logins(issue)
                     if current_assignees:
-                        if self._user_in(pr_author, current_assignees):
+                        if user_in_logins(pr_author, current_assignees):
                             print(
                                 f"Issue #{issue_number} already"
                                 f" assigned to {pr_author}"
@@ -365,7 +362,7 @@ class IssueAssignmentBot(GitHubBot):
                             )
                         continue
                     issue.add_to_assignees(pr_author)
-                    verified = self._assignment_succeeded(issue_number, pr_author)
+                    verified = self._verify_assignment(issue_number, pr_author)
                     if verified is True:
                         assigned_issues.append(issue_number)
                         print(f"Assigned issue #{issue_number} to {pr_author}")
@@ -379,18 +376,18 @@ class IssueAssignmentBot(GitHubBot):
                     elif verified is False:
                         print(
                             f"Assignment of #{issue_number} to {pr_author}"
-                            " was silently rejected by GitHub"
+                            " was silently rejected by GitHub."
                         )
                         issue.create_comment(
-                            self._build_silent_failure_message(pr_author, pr_number)
+                            self._cannot_auto_assign_message(pr_author, pr_number)
                         )
                     else:
                         print(
                             f"Skipping comment for #{issue_number}:"
-                            " assignment state could not be verified"
+                            " assignment state could not be verified."
                         )
                 except Exception as e:
-                    print(f"Error processing issue" f" #{issue_number}: {e}")
+                    print(f"Error processing issue #{issue_number}: {e}")
             return assigned_issues
         except Exception as e:
             print(f"Error in auto_assign_issues_from_pr: {e}")
--- a/.github/actions/bot-autoassign/issue_assignment_bot.py
+++ b/.github/actions/bot-autoassign/issue_assignment_bot.py
@@ -1,14 +1,30 @@
+import os
 import re
+import time
 
 from base import GitHubBot
+from github import GithubException
 from utils import (
     extract_linked_issues,
+    find_open_pr_for_issue,
     get_valid_linked_issues,
     unassign_linked_issues_helper,
 )
 
+VERIFICATION_RETRY_DELAY_SECONDS = 1.0
+
 
 class IssueAssignmentBot(GitHubBot):
+    def __init__(self):
+        super().__init__()
+        self.bot_username = os.environ.get("BOT_USERNAME", "openwisp-companion")
+
+    def is_bot_assign_command(self, comment_body):
+        if not comment_body:
+            return False
+        pattern = rf"(?<![\w-])@{re.escape(self.bot_username)}\s+assign\b"
+        return bool(re.search(pattern, comment_body, re.IGNORECASE))
+
     def is_assignment_request(self, comment_body):
         if not comment_body:
             return False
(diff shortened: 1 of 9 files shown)


nemesifier:
@coderabbitai review

nemesifier:
@coderabbitai resolve

.github/actions/bot-autoassign/tests/test_issue_assignment_bot.py
```diff
@@ -103,6 +104,24 @@ def test_extract_linked_issues(self, pr_body, expected, bot_env):
...
+            ("Closes #456 and resolves #789", [456, 789]),
+            ("Fixed #999", [999]),
+            ("Related to #99", []),  # excluded in strict mode
+            ("Relates to #42", []),  # excluded in strict mode
+            ("Fixes #1 and related to #2", [1]),
+            ("", []),
+        ],
+    )
+    def test_extract_linked_issues_strict(self, pr_body, expected, bot_env):
+        from utils import extract_linked_issues
```
nemesifier: move import at the top

.github/actions/bot-autoassign/issue_assignment_bot.py
```diff
@@ -194,6 +210,118 @@ def respond_to_assignment_request(self, issue_number, commenter):
...
+                    f"GitHub API error verifying assignment for #{issue_number}"
+                    f" (attempt {attempt + 1}/{attempts}): {e}"
+                )
+            if attempt + 1 < attempts:
+                time.sleep(VERIFICATION_RETRY_DELAY_SECONDS)
+        if last_error is not None:
+            return None
+        return False
+
+    def _build_silent_failure_message(self, pr_author, pr_number):
```
nemesifier: Let's use more explicit names please:

```suggestion
    def _cannot_auto_assign_message(self, pr_author, pr_number):
```

.github/actions/bot-autoassign/issue_assignment_bot.py
```diff
@@ -194,6 +210,118 @@ def respond_to_assignment_request(self, issue_number, commenter):
...
+                print(f"{commenter} is already assigned to #{issue_number}")
+                return True
+            pr = find_open_pr_for_issue(self.repo, commenter, issue_number)
+            if pr is None:
+                issue.create_comment(
+                    f"Hi @{commenter} 👋,\n\n"
+                    "I could not find an open PR by you that references"
+                    f" this issue (#{issue_number}). Please open a PR"
+                    f" linking to this issue (e.g. `Fixes #{issue_number}`)"
+                    " and then re-run the command."
```
nemesifier: Which command?

.github/actions/bot-autoassign/issue_assignment_bot.py
```diff
@@ -194,6 +210,118 @@ def respond_to_assignment_request(self, issue_number, commenter):
...
+                issue.create_comment(
+                    f"This issue has been assigned to @{commenter}"
+                    f" who opened PR #{pr.number} to address it. 🎯"
+                )
+                print(f"Assigned #{issue_number} to {commenter} via bot command")
+            elif verified is False:
+                issue.create_comment(
+                    f"Sorry @{commenter}, GitHub still did not allow"
+                    " this assignment. A maintainer will need to assign"
+                    " this issue manually."
```
nemesifier, line 280: in what case/cases would this happen? A comment that summarizes this would be good.

.github/actions/bot-autoassign/issue_assignment_bot.py
```diff
@@ -194,6 +210,118 @@ def respond_to_assignment_request(self, issue_number, commenter):
...
+                )
+                print(
+                    f"Bot-command assignment of #{issue_number} to"
+                    f" {commenter} was silently rejected"
```
nemesifier: For consistency:
```suggestion
                    f" {commenter} was silently rejected."
```

.github/actions/bot-autoassign/issue_assignment_bot.py
```diff
@@ -194,6 +210,118 @@ def respond_to_assignment_request(self, issue_number, commenter):
...
+                )
+            else:
+                print(
+                    f"Skipping comment for #{issue_number}:"
+                    " assignment state could not be verified"
```
nemesifier: For consistency:
```suggestion
                    " assignment state could not be verified."
```

nemesifier (changes requested):

nemesifier (changes requested):
### `last_error` not cleared after a successful fetch in `_assignment_succeeded`

`issue_assignment_bot.py:_assignment_succeeded`

The loop runs up to two attempts. If attempt 0 raises `GithubException` (`last_error` is set),
then attempt 1 calls `get_issue` successfully but the user is not yet in the assignee list, the
loop exits with `last_error is not None`. The trailing guard then returns `None` (silent, no
comment posted) instead of `False` (assignment verifiably failed, post the fallback message). The
fix is to reset `last_error = None` immediately after `self.repo.get_issue(issue_number)` returns
without raising.

### Non-`GithubException` errors escape `_assignment_succeeded`'s contract

`issue_assignment_bot.py:_assignment_succeeded`

The method's contract states it returns `None` when verification itself errors so the caller stays
silent. Only `GithubException` is caught; lower-level network or serialization errors bubble up to
the outer `except Exception` in the callers, which return `False` for the whole operation rather
than staying silent. Catch `Exception` (or at minimum add a catch-all `except Exception` after
`GithubException`) to honour the stated contract.

### `strict=True` breaks the fallback flow for "Related to" PRs

`utils.py:find_open_pr_for_issue` / `issue_assignment_bot.py:handle_bot_assign_request`

`auto_assign_issues_from_pr` uses the default (non-strict) `extract_linked_issues`, so a PR with
`Related to #123` triggers auto-assignment of issue 123. If that assignment is silently rejected,
`_build_silent_failure_message` is posted, telling the contributor to comment
`@openwisp-companion assign`. When they do, `handle_bot_assign_request` calls
`find_open_pr_for_issue` with `strict=True`, which excludes "Related to" links, and the bot
replies "I could not find an open PR by you". The user followed the bot's own instructions and
still cannot get assigned.

The call site in `handle_bot_assign_request` should pass `strict=False` (or rely on the default)
to stay consistent with the auto-assign path.

### `unassign_linked_issues_helper` uses case-sensitive comparison inconsistency

`utils.py:unassign_linked_issues_helper`

This PR introduces `_user_in` for case-insensitive login comparison and applies it throughout
`IssueAssignmentBot`. `unassign_linked_issues_helper` still does `pr_author in current_assignees`
with a plain string comparison. GitHub logins are case-insensitive, so if the casing returned by
the API differs from what was stored, the unassign step silently does nothing. The helper should
use the same lowercase-both-sides logic.

.github/actions/bot-autoassign/tests/test_issue_assignment_bot.py
Eeshu-Yadav: Moved to the top.

.github/actions/bot-autoassign/issue_assignment_bot.py
Eeshu-Yadav: Renamed.

.github/actions/bot-autoassign/issue_assignment_bot.py
Eeshu-Yadav: Clarified — the message now spells out `@openwisp-companion assign` instead of "the command".

.github/actions/bot-autoassign/issue_assignment_bot.py
Eeshu-Yadav: Added a short comment — this hits when the commenter has satisfied GitHub's "must have commented" rule but the assignment still fails (e.g. a repo-level permission block or an API outage).

.github/actions/bot-autoassign/issue_assignment_bot.py
Eeshu-Yadav: Done.

.github/actions/bot-autoassign/issue_assignment_bot.py
Eeshu-Yadav: Done.

Eeshu-Yadav:
@coderabbitai review

nemesifier (changes requested):
@Eeshu-Yadav great progress!

The testing confirms the full flow works end-to-end, excellent! 👍

See my comments below to polish the code a bit more so we can merge.

### Misleading variable name in `_assignment_succeeded`

`issue_assignment_bot.py:218-232`

`errored_all` starts as `True` and is set to `False` the first time a `get_issue` call returns successfully. A reader has to hold that inversion in their head throughout the loop. The intended meaning is "has at least one fetch succeeded", so the variable reads backwards relative to its name. A clearer formulation:

```python
def _assignment_succeeded(self, issue_number, user):
    had_successful_fetch = False
    for attempt in range(2):
        try:
            updated_issue = self.repo.get_issue(issue_number)
            had_successful_fetch = True
            if user_in_logins(user, self._get_assignee_logins(updated_issue)):
                return True
        except Exception as e:
            print(...)
        if attempt == 0:
            time.sleep(VERIFICATION_RETRY_DELAY_SECONDS)
    return False if had_successful_fetch else None
```

### Method name `_assignment_succeeded` implies a boolean but can return None

`issue_assignment_bot.py:217`

The method returns `True`, `False`, or `None`. A name ending in "succeeded" implies a bool, which is what callers see in `if verified is True` / `elif verified is False` / `else`. Something like `_verify_assignment` signals that the result must be compared explicitly rather than used as a plain truthiness check. Not urgent, but will trip up anyone reading the code cold.

Is `None` intentional here? If yes, please add a brief docstring explaining what's the difference between `None` and `False`.

### `find_open_pr_for_issue` only catches `GithubException`

`utils.py:68`

```python
except GithubException as e:
    print(f"GitHub API error searching open PRs by {author}: {e}")
return None
```

`_assignment_succeeded` catches `Exception` for consistency with the "stay silent on unexpected errors" contract. `find_open_pr_for_issue` does not. If a lower-level network error (e.g. `requests.exceptions.ConnectionError`) is raised during the search, it propagates to the outer `except Exception` in `handle_bot_assign_request`, which logs "Error handling bot assign command" and returns `False` — and the user gets no comment at all. Catching `Exception` here would let the function return `None` gracefully and keep the caller's logic intact.

### `extract_linked_issues` is called twice on the same PR body

`issue_assignment_bot.py:318` and `utils.py:30`

`auto_assign_issues_from_pr` calls `extract_linked_issues(pr_body)` to check for an early exit and to log the rate-limit warning, then immediately calls `get_valid_linked_issues(…, pr_body)` which calls `extract_linked_issues` again internally. The PR body is typically small so the overhead is negligible, but two calls on the same input is fragile: anyone who later makes `extract_linked_issues` stateful or expensive will have to trace this dependency. The simpler fix is to call it once, store the result, and pass the list into `get_valid_linked_issues` (or skip the outer call and use `get_valid_linked_issues`'s result for the no-issues check).

### Redundant f-string split in `auto_assign_issues_from_pr`

`issue_assignment_bot.py:381`

```python
print(f"Error processing issue" f" #{issue_number}: {e}")
```

Two adjacent f-strings with no reason to be separate. Should be one:

```python
print(f"Error processing issue #{issue_number}: {e}")
```

### Missing test: fetch succeeds on attempt 0 (no assignee), fails on attempt 1

The test suite covers:
- both fetches fail → `None` (`test_verification_error_stays_silent`)
- fetch 0 fails, fetch 1 succeeds with empty assignees → `False` (`test_verification_recovers_then_reports_failure`)
- fetch 0 stale, fetch 1 fresh with assignee → `True` (`test_verification_retries_on_transient_lag`)

The untested case is: fetch 0 succeeds (user not in assignees, `errored_all = False`), fetch 1 raises. The return is `False` because `had_successful_fetch` (or `not errored_all`) is True. This is the correct behaviour — attempt 0 gave reliable evidence the user is not assigned — but there is no test for it.

### `reusable-bot-autoassign.yml` still points to the testing fork

`reusable-bot-autoassign.yml:35-37`

```yaml
# TEMPORARY: testing #649 fix — revert to openwisp/openwisp-utils@master before merge.
repository: Eeshu-Yadav/openwisp-utils
ref: issues/649-fix-bot-assignment-silent-failure
```

This must be reverted to `repository: openwisp/openwisp-utils` / `ref: master` before the PR is merged. Any downstream repository that calls the reusable workflow will continue pulling from the contributor's fork after merge, which means their bots will silently break the moment the fork diverges or is deleted.

nemesifier (changes requested):
@Eeshu-Yadav great progress! All the points from my previous review (variable name, method rename + docstring, broader exception handling, the duplicate `extract_linked_issues` call, the f-string split, the missing test for "fetch 0 succeeds, fetch 1 raises", and the workflow ref pointing to your fork) have been addressed in 3240f80. The test suite is in good shape.

Two small items left before we can merge.

### `pr_reopen_bot.py` still uses case-sensitive login comparisons

The fixup nicely extracted `user_in_logins` and applied it to `unassign_linked_issues_helper`, but the same pattern is still present in `pr_reopen_bot.py`:

`pr_reopen_bot.py:22` and `pr_reopen_bot.py:29` (in `PRReopenBot.reassign_issues_to_author`):

```python
if current_assignees and pr_author not in current_assignees:
    ...
if pr_author not in current_assignees:
    issue.add_to_assignees(pr_author)
```

`pr_reopen_bot.py:119` (in `PRActivityBot.handle_contributor_activity`):

```python
if not pr.user or commenter != pr.user.login:
```

These are pre-existing on master, but since `user_in_logins` now exists in `utils.py` and we've just fixed the same silent-no-op bug in the unassign helper for the same reason, please apply it consistently here too. Otherwise the casing-mismatch silent no-op still exists on the reopen / stale-PR flows.

### Docs: the new `bot_username` reusable-workflow input is undocumented

`reusable-bot-autoassign.yml` now exposes a `bot_username` input (defaulting to `openwisp-companion`), but `docs/developer/reusable-github-utils.rst` doesn't mention it. Downstream OpenWISP repositories using a different bot username will have no way to discover that they can override the default. A short note in the existing reusable-bot-autoassign section is enough, one sentence about the input plus an example caller passing `bot_username:`. The docs checkbox in the PR description is also still unticked, please update it once the rst note is in.

### (Optional) CodeRabbit's `BOT_USERNAME` test suggestion

CodeRabbit suggested one end-to-end test exercising a non-default `BOT_USERNAME` so the configurable workflow path stays covered. Not blocking, but nice to have if it's quick.

After these are in, this should be ready to merge. Thanks again for the thoroughness on this fix, the end-to-end testing on `bot-testing-ground` was excellent! 👍

nemesifier (commented):
@Eeshu-Yadav pushed the remaining changes, please double check when you can.

docs/developer/reusable-github-utils.rst
```diff
@@ -218,6 +218,26 @@ Create the following workflow files in your repository.
           OPENWISP_BOT_APP_ID: ${{ secrets.OPENWISP_BOT_APP_ID }}
           OPENWISP_BOT_PRIVATE_KEY: ${{ secrets.OPENWISP_BOT_PRIVATE_KEY }}
 
+**Overriding the bot username**
+
+All four caller workflows accept an optional ``bot_username`` input
+(default ``openwisp-companion``). The bot uses it to detect mentions of
+the form ``@<bot_username> assign`` in issue comments and to ignore
+comments authored by the bot itself. Set this if your repository uses a
+different GitHub App username:
```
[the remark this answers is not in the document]
Eeshu-Yadav: Skipping — the override section right above already documents the `bot_username` workflow input, default value, and example. Since the repo only uses one bot (`openwisp-companion`) and that section covers the rare downstream-fork override case, adding `BOT_USERNAME` to the secrets-paragraph runtime env-var list would just clutter without helping a reader.

Eeshu-Yadav:
Two follow-up findings :

**1. Same silent-rejection bug class still exists in the reassign paths.** The fix in `auto_assign_issues_from_pr` doesn't cover:

- `pr_reopen_bot.py:31-41` (`PRReopenBot.reassign_issues_to_author`) — `issue.add_to_assignees(pr_author)` followed by `reassigned_issues.append(...)` + welcome comment, no verification.
- `pr_reopen_bot.py:144-157` (`PRActivityBot.handle_contributor_activity`) — `issue.add_to_assignees(commenter)` followed by `reassigned_count += 1` + encouragement comment, no verification.

`PRActivityBot` is the sharper trap: it fires on a PR comment, which doesn't satisfy GitHub's "must have commented on the linked issue" rule, so a non-collaborator poking a stale PR can hit silent rejection while the bot still claims the linked issue was reassigned. Fix is mechanical — lift `_verify_assignment` into `utils.py` and gate the bookkeeping + comment on `verified is True` in both call sites.

**2. `bot-autoassign-pr-issue-link.yml` concurrency cancels in-flight close-handler.**

```yaml
concurrency:
  group: bot-autoassign-pr-link-${{ github.repository }}-${{ github.event.pull_request.number }}
  cancel-in-progress: true
```

Workflow runs on `[opened, reopened, closed]`; a fast open→close (or close→reopen) cancels the prior run, leaving assignment state inconsistent (e.g., a `closed`-event unassign killed by an immediately following `reopened`). Same pattern is mirrored in the docs example. Fix: append `-${{ github.event.action }}` to the group key so each action type serializes independently.

Will pick these up in a follow-up.

nemesifier:
@coderabbitai full review

nemesifier (approved):

nemesifier:
@coderabbitai resolve

PR #655, opened 2026-04-24
[chores:fix] Removed 'FAILED (' from strict markers to unblock auto-retry

stktyagi:
Removed 'FAILED' keyword from strict failure markers which caused blocking of re-run mechanism

## Checklist

## Description of Changes

Existence of 'FAILED' keyword in strict failure markers blocked re-run mechanism due its occurence in any case of failure in CI failure logs.

### DIFF
--- a/.github/actions/bot-ci-failure/analyze_failure.py
+++ b/.github/actions/bot-ci-failure/analyze_failure.py
@@ -9,7 +9,6 @@ from google.genai import types
 # Strict markers that unequivocally indicate a failing test or assertion.
 STRICT_TEST_FAILURE_MARKERS = (
     "FAIL:",
-    "FAILED (",
     "AssertionError",
 )
 
--- a/.github/actions/bot-ci-failure/analyze_failure.py
+++ b/.github/actions/bot-ci-failure/analyze_failure.py
@@ -34,7 +34,7 @@ TRANSIENT_FAILURE_MARKERS = (
     "Network is unreachable",
     "Temporary failure in name resolution",
     "selenium.common.exceptions.InvalidSessionIdException",
-    "selenium.common.exceptions.WebDriverException"
+    "selenium.common.exceptions.WebDriverException",
     "Posting coverage data to https://coveralls.io",
     "OperationalError: database is locked",
     "ERROR: Could not install packages due to an OSError",
--- a/.github/actions/bot-ci-failure/analyze_failure.py
+++ b/.github/actions/bot-ci-failure/analyze_failure.py
@@ -34,6 +34,7 @@ TRANSIENT_FAILURE_MARKERS = (
     "Network is unreachable",
     "Temporary failure in name resolution",
     "selenium.common.exceptions.InvalidSessionIdException",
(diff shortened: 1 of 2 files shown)


stktyagi:
@coderabbitai full review

stktyagi:
@coderabbitai resolve

nemesifier (commented):
@stktyagi I am thinking.. we are trying to figure out which of these 3 cases we're in:

1. tests failed due to flakyness but also has real test failures
2. tests failed only due to flakyness
3. tests failed only due to real test failures

I think the problem is mainly discern between 1 and 2. Implementing this with text patterns and make it work across different programming languages (python, lua, nodejs, etc) will be super tricky.

Maybe we can make it work for Python, but I am skeptical this will work also for the nodejs repos.

Are you sure this is the right approach?
Why don't we give all the context to the LLM and ask it to tell us if we are under case 1 or 2? The LLM should be good at this. We could do this in a separate API request, which has the sole goal of understanding if we are falling under case 1 or case 2, based on the result of the response we'll either restart the build or not, what do you think of this?

stktyagi:
> what do you think of this?

My main concern with that approach would be handling hallucinations, we would need to be deterministic with whether a CI failure is due to actual test failure or flakyness. Also, while Node.js and Python format tracebacks differently, the actual infrastructure crashes we want to forgive are mostly consistent across all CIs and even if we encounter new ones, I feel including them iteratively until we reach max coverage seems fine.

It feels like a trade-off :- Non-deterministic approach would be language versatile but might hallucinate, whereas current approach would be deterministic but we would need to consistently monitor new transient error inclusions and iterations. Depends which seems better to us.

nemesifier:
> > what do you think of this?
> 
> My main concern with that approach would be handling hallucinations, we would need to be deterministic with whether a CI failure is due to actual test failure or flakyness. Also, while Node.js and Python format tracebacks differently, the actual infrastructure crashes we want to forgive are mostly consistent across all CIs and even if we encounter new ones, I feel including them iteratively until we reach max coverage seems fine.
> 
> It feels like a trade-off :- Non-deterministic approach would be language versatile but might hallucinate, whereas current approach would be deterministic but we would need to consistently monitor new transient error inclusions and iterations. Depends which seems better to us.

I understand, but we spent a lot of time on this and we don't have something working. If this continues I'd rather drop this feature entirely as it's not sustainable to keep working on something with no results. This retry feature has never worked, it's a nice idea I wanted and I'd love to have it, but I am not willing to spend hundred of hours on this, we have other priorities.

stktyagi:
> This retry feature has never worked, it's a nice idea I wanted and I'd love to have it, but I am not willing to spend hundred of hours on this, we have other priorities.

That's valid. In that case we can hand the decision to an LLM of whether it is a transient error or not. Either way, the worst that could happen is that it hallucinates a genuine error once in a while and re-runs the CI 3 times.
Let me know how you want to proceed.

nemesifier (approved):
Let's see how this change goes and then we can reassess.

PR #662, opened 2026-05-02
 System info improvements #376

### LINKED ISSUE #376: [change] System info improvements
We should add the possibility to specify permissions to see the menu links, I think this only works for model links now.
We may allow only superusers to see the system info or we may create a specific permission.

Let's add the license of OpenWISP in the system info page.

BHARATH0153:
Added superuser-only permission to the system info page and menu link. Non-superuser staff accounts now receive a 403 when accessing the URL and the System info menu entry is hidden from them entirely.

Added OpenWISP license display in the system info page by reading the LICENSE file from the package root and passing it to the template.

Added permissions support to MenuLink so that menu items can declare superuser_only restriction. The system info menu entry in apps.py uses this new permissions key.

Closes #376

### DIFF
--- a/openwisp_utils/admin_theme/admin.py
+++ b/openwisp_utils/admin_theme/admin.py
@@ -12,6 +12,7 @@ from . import settings as app_settings
 from .dashboard import get_dashboard_context
 from .system_info import (
     get_enabled_openwisp_modules,
+    get_openwisp_license,
     get_openwisp_version,
     get_os_details,
 )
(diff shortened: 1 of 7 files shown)


nemesifier:
This is wrong, stop wasting my time, I am tired of babysitting people who can't even read the issue discussion. I spent months reviewing PRs like this, now this is over, STOP IT! Discuss with maintainers the expeted outcomes before raising PRs!

PR #666, opened 2026-05-06
 [fix] Fixed double encoding of device model labels in dashboard #651 

### LINKED ISSUE #651: [bug] Double Encoding in Admin Dashboard Device Models segment
**Clicking on dashboard Device Models pie chart segments generates incorrect filter URLs when labels contain special characters such as "&".**

**Steps To Reproduce**
Device model:
Dongwon T&I DW02-412H (128M)

In Admin dashboard  Network Administration
`Device Models` chart
clicking on this device filter 

Generated URL:
`/admin/config/device/?model__exact=Dongwon+T%26amp%3BI+DW02-412H+%28128M%29`

which is double encoded

**Expected behavior**
Correct URL:
`/admin/config/device/?model__exact=Dongwon%20T%26I%20DW02-412H%20%28128M%29`

`System type` chart generates correct html

**System Information:**

- OS: [Ubuntu 22 LTS]
- Python Version: [3.10.12]
- Django Version: [Django 5.2.11]
- Browser and Browser Version: [Chrome v147.0.7727.57]

I have changed this 

https://github.com/openwisp/openwisp-utils/blob/f6305c424d34364fb321d759f7d7faa69ddccb42/openwisp_utils/admin_theme/dashboard.py#L221

to `label = (label)`

works for me, i do not know if this is the correct approach or not.

BHARATH0153:
## Checklist

 I have read the OpenWISP Contributing Guidelines.

I have manually tested the changes proposed in this pull request.

 I have written new test cases for new code and/or updated existing tests for changes to existing code.

 I have updated the documentation.

Fixes #651 

## Description of Changes
Hoisted filters.append Moved the append logic above the if/else block to remove the duplication and keep the code DRY.

Server-side URL Encoding Wrapped the label in quote(label, safe="") from urllib.parse before appending it to filters. This ensures the JS click handler receives a fully URL-encoded string (e.g., %26 instead of &), preventing the URL parsing issue in the browser.

## Local validation:
black openwisp_utils/admin_theme/dashboard.py tests/test_project/tests/test_dashboard.py
python runtests.py test_project.tests.test_dashboard.TestAdminDashboard.test_filter_not_double_encoded -v 2
python runtests.py test_project.tests.test_dashboard.TestAdminDashboard.test_index_content -v 2

## Screenshot

f1
before fix 

after fix
after fix

### DIFF
--- a/openwisp_utils/admin_theme/dashboard.py
+++ b/openwisp_utils/admin_theme/dashboard.py
@@ -1,5 +1,6 @@
 import copy
 import html
+from urllib.parse import quote
 
 from django.core.exceptions import ImproperlyConfigured
 from django.db.models import Count
@@ -209,10 +210,8 @@ def get_dashboard_context(request):
                 label = qs_key
                 # get human readable label if predefined labels are available
                 # otherwise use the result got from the DB
+                filters.append(quote(label, safe=""))
                 if labels_i18n and qs_key in labels_i18n:
-                    # store original label as filter, but only
-                    # if we have more than the empty default label defined
-                    filters.append(label)
                     label = labels_i18n[qs_key]
                 else:
                     # HTML escape labels coming from values in the DB
--- a/openwisp_utils/admin_theme/dashboard.py
+++ b/openwisp_utils/admin_theme/dashboard.py
@@ -208,9 +208,10 @@ def get_dashboard_context(request):
                     continue
                 qs_key = str(obj[group_by])
                 label = qs_key
+                # add URL quoted label to filters
+                filters.append(quote(label, safe=""))
                 # get human readable label if predefined labels are available
                 # otherwise use the result got from the DB
-                filters.append(quote(label, safe=""))
                 if labels_i18n and qs_key in labels_i18n:
                     label = labels_i18n[qs_key]
                 else:
--- a/tests/test_project/tests/test_dashboard.py
+++ b/tests/test_project/tests/test_dashboard.py
@@ -294,3 +294,30 @@ class TestAdminDashboard(AdminTestMixin, CreateMixin, DjangoTestCase):
             context["dashboard_charts"][1]["query_params"]["labels"][0],
             "&lt;strong&gt;Projects with operators&lt;/strong&gt;",
         )
+
+    def test_filter_not_double_encoded(self):
+        project = Project.objects.create(name="Dongwon T&I")
+        Operator.objects.create(project=project, first_name="test", last_name="test")
+        mocked_user = MockUser(is_superuser=True)
+        mocked_request = MockRequest(user=mocked_user)
+        context = get_dashboard_context(mocked_request)
+        target_chart = None
+        for chart in context["dashboard_charts"].values():
+            if chart.get("name") == "Operator Project Distribution":
+                target_chart = chart
+                break
+        self.assertIsNotNone(target_chart)
+        filters = target_chart.get("filters", [])
+        labels = target_chart.get("query_params", {}).get("labels", [])
+        self.assertIn("Dongwon%20T%26I", filters)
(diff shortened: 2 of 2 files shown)


BHARATH0153:
@coderabbitai review

BHARATH0153:
@nemesifier please review when you are free
locally reproduced  and fixed  as above mentioned the proofs https://github.com/openwisp/openwisp-utils/pull/666#issue-4391942796 (screenshot) thanks!

BHARATH0153:
@coderabbitai check the written new test in test_dashboard.py and review the changes is ready to merge

BHARATH0153:
@coderabbitai review

BHARATH0153:
@coderabbitai check the written new test in test_dashboard.py and review the changes is ready to merge

BHARATH0153:
@coderabbitai review

BHARATH0153:
@coderabbitai

nemesifier (changes requested):
The fix lands the right server-side split, keep the raw label for the filter, escape it for display, but two things still need attention before this can merge.

`filters.append(label)` ends up in both branches of the if/else:

```python
if labels_i18n and qs_key in labels_i18n:
    filters.append(label)
    label = labels_i18n[qs_key]
else:
    filters.append(label)
    # HTML escape labels coming from values in the DB
    # to avoid possible XSS attacks caused by
    # malicious DB values set by users
    label = html.escape(label)
```

Same line in both branches always wants to be hoisted above the if/else. Doing that also fixes a small reading hiccup: the new `filters.append(label)` currently sits between `else:` and the comment about `html.escape`, so the comment no longer describes the line right below it.

```python
filters.append(label)
if labels_i18n and qs_key in labels_i18n:
    label = labels_i18n[qs_key]
else:
    # HTML escape labels coming from values in the DB
    # to avoid possible XSS attacks caused by
    # malicious DB values set by users
    label = html.escape(label)
```

The bigger problem is that the JS this fix relies on never URL-encodes what it appends. From `ow-dashboard.js`:

```js
if (filters && typeof filters[i] !== "undefined") {
    path += filters[i];
}
```

After this PR the click handler receives the raw `Dongwon T&I DW02-412H (128M)` and concatenates it onto the path. Browsers don't auto-encode `&`, `=`, `?`, `#`, `(`, `)`, so the URL that ends up at the server is `?model__exact=Dongwon T&I DW02-412H (128M)`, which parses at the `&` into two query parameters and applies the filter as `model__exact=Dongwon T`. The expected URL in #651 (`?model__exact=Dongwon%20T%26I%20DW02-412H%20%28128M%29`) needs `encodeURIComponent` to land, either on the JS side (`path += encodeURIComponent(filters[i])`) or by pre-encoding the values when the server populates `filters`.

The new test doesn't surface this because it only checks what the server writes into `chart["filters"]` and `chart["query_params"]["labels"]`; it doesn't build the URL the JS would actually navigate to. The double-encoding symptom is gone, but for the exact labels that originally hit this bug the resulting URL is malformed instead. Worth extending the coverage to the click handler's output, Selenium, or a small unit test that replicates the string concatenation, and either fixing the encoding in JS or pre-encoding on the server.

BHARATH0153:
@coderabbitai  is it ready to merge now after maintainer review

BHARATH0153:
@nemesifier  oooooo
 changes are addressed and tests are passing

BHARATH0153:
@coderabbitai check overall improvement and review and is it ready to merge and check fulfillment of @nemesifier changes or comment

BHARATH0153:
@coderabbitai review

BHARATH0153:
@coderabbitai now is it ready for merge check new commit

nemesifier (approved):

PR #668, opened 2026-05-08
[ci:change] Stale-PR bot: drop auto-close and fix false-positive detection

Eeshu-Yadav:
## Checklist

Fixes the stale-PR bot incident where ~16 PRs across `openwisp-utils`, `openwisp-controller`, and `openwisp-monitoring` were auto-closed after the bot mistook bot reviews (coderabbitai) for blocking maintainer reviews. Federico's temporary stub on master disabled auto-close while this proper fix was prepared.

## What changes

**Detection (the actual bug)**
- `get_last_changes_requested` now filters out bot reviews and tracks each reviewer's latest non-COMMENTED state — a reviewer who later approves or dismisses no longer blocks.
- `_commit_activity_date_for_author` picks the date of whichever identity (author or committer) actually matches the PR author, so a maintainer rebasing the contributor's commits is no longer counted as contributor activity.

**Auto-close removed**
- `close_stale_pr` is gone. At 60 days the bot now posts a `final_followup` encouragement comment and leaves the PR open. Maintainers can close manually if a PR is genuinely abandoned.

**Stale-state recovery**
- When the cron sees contributor activity past `last_changes_requested`, the bot clears the `stale` label and reassigns linked issues that are currently unassigned. Previously the label and issue state could persist indefinitely.

**Tighter `is_waiting_for_maintainer`**
- Only counts maintainer **reviews**, not comments — a follow-up question in a comment doesn't hand the ball back to the contributor.
- Fail-closed: on any exception, the bot now skips the PR instead of proceeding.

**Other**
- `has_bot_comment` matches our bot's identity so a third-party bot reusing the marker can't suppress our stages.
- `pr.get_commits()` is materialized once per PR and threaded through, removing duplicate API iteration.
- `mark_pr_stale` unassigns issues before posting the "unassigned" comment, so the comment can't claim something that didn't happen.
- Final-followup waits a run after the stale notification so the two comments don't land back-to-back.

## Testing on `bot-testing-ground`

| Scenario | Reference |
|----------|-----------|
| Bot-only CR ignored | [bot-testing-ground#103](https://github.com/openwisp/bot-testing-ground/pull/103) |
| Stale at 14 days (label + unassign) | [bot-testing-ground#102](https://github.com/openwisp/bot-testing-ground/pull/102)  |
| Stale-label recovery on contributor push | [bot-testing-ground#102](https://github.com/openwisp/bot-testing-ground/pull/102)  |
| Day-60 final follow-up, PR stays open | [bot-testing-ground#102](https://github.com/openwisp/bot-testing-ground/pull/102)  |
| Supersession by APPROVED | [bot-testing-ground#104](https://github.com/openwisp/bot-testing-ground/pull/104) |
| Supersession by DISMISSED | [bot-testing-ground#105](https://github.com/openwisp/bot-testing-ground/pull/105) |
| Maintainer comment doesn't reset | [bot-testing-ground#106](https://github.com/openwisp/bot-testing-ground/pull/106) |
| Third-party marker not trusted | [bot-testing-ground#107](https://github.com/openwisp/bot-testing-ground/pull/107) |
| First-time PR past 60 days | [bot-testing-ground#108](https://github.com/openwisp/bot-testing-ground/pull/108) |
| Maintainer rebase doesn't reset clock | [bot-testing-ground#109](https://github.com/openwisp/bot-testing-ground/pull/109) |

### DIFF
--- a/.github/actions/bot-autoassign/stale_pr_bot.py
+++ b/.github/actions/bot-autoassign/stale_pr_bot.py
@@ -32,11 +32,21 @@ class StalePRBot(GitHubBot):
             return None
         last_activity = None
         for commit in pr.get_commits():
-            commit_date = commit.commit.author.date
-            if commit_date > after_date:
-                if commit.author and commit.author.login == pr_author:
-                    if not last_activity or commit_date > last_activity:
-                        last_activity = commit_date
+            author_date = commit.commit.author.date if commit.commit.author else None
+            committer_date = (
+                commit.commit.committer.date if commit.commit.committer else None
+            )
+            if author_date and committer_date:
+                commit_date = max(author_date, committer_date)
+            else:
+                commit_date = author_date or committer_date
+            if not commit_date or commit_date <= after_date:
+                continue
+            author_login = commit.author.login if commit.author else None
+            committer_login = commit.committer.login if commit.committer else None
+            if author_login == pr_author or committer_login == pr_author:
+                if not last_activity or commit_date > last_activity:
+                    last_activity = commit_date
         if issue_comments is None:
             issue_comments = list(pr.get_issue_comments())
         for comment in issue_comments:
@@ -160,13 +170,25 @@ class StalePRBot(GitHubBot):
         try:
             if all_reviews is None:
                 all_reviews = list(pr.get_reviews())
-            changes_requested_reviews = [
-                r for r in all_reviews if r.state == "CHANGES_REQUESTED"
+            latest_per_reviewer = {}
+            for r in all_reviews:
+                if not r.user or not r.submitted_at:
+                    continue
+                if r.user.type == "Bot":
+                    continue
+                if r.state == "COMMENTED":
+                    continue
+                current = latest_per_reviewer.get(r.user.login)
+                if current is None or r.submitted_at > current.submitted_at:
+                    latest_per_reviewer[r.user.login] = r
+            blocking = [
+                r
+                for r in latest_per_reviewer.values()
+                if r.state == "CHANGES_REQUESTED"
             ]
-            if not changes_requested_reviews:
+            if not blocking:
                 return None
-            changes_requested_reviews.sort(key=lambda r: r.submitted_at, reverse=True)
-            return changes_requested_reviews[0].submitted_at
+            return max(blocking, key=lambda r: r.submitted_at).submitted_at
         except Exception as e:
             print("Error getting reviews" f" for PR #{pr.number}: {e}")
             return None
--- a/.github/actions/bot-autoassign/stale_pr_bot.py
+++ b/.github/actions/bot-autoassign/stale_pr_bot.py
@@ -19,7 +19,10 @@ class StalePRBot(GitHubBot):
         self.DAYS_BEFORE_STALE_WARNING = 7
         self.DAYS_BEFORE_UNASSIGN = 14
         self.DAYS_BEFORE_FINAL_FOLLOWUP = 60
-        self.bot_login = os.environ.get("BOT_USERNAME", "openwisp-companion") + "[bot]"
+        bot_username = os.environ.get("BOT_USERNAME", "openwisp-companion")
+        self.bot_login = (
+            bot_username if bot_username.endswith("[bot]") else f"{bot_username}[bot]"
+        )
 
     @staticmethod
     def _commit_activity_date_for_author(commit, pr_author):
@@ -164,34 +167,33 @@ class StalePRBot(GitHubBot):
     def get_last_changes_requested(self, pr, all_reviews=None):
         """Timestamp of the latest CHANGES_REQUESTED that still represents
         a human reviewer's current stance, or ``None``.
+
+        Errors propagate so the caller can distinguish "no active block"
+        from "couldn't determine" and skip the PR.
         """
-        try:
-            if all_reviews is None:
-                all_reviews = list(pr.get_reviews())
-            # Bot reviews are advisory; COMMENTED does not change stance.
-            latest_per_reviewer = {}
-            for review in all_reviews:
-                if (
-                    not review.user
-                    or not review.submitted_at
-                    or review.user.type == "Bot"
-                    or review.state == "COMMENTED"
-                ):
-                    continue
-                current = latest_per_reviewer.get(review.user.login)
-                if current is None or review.submitted_at > current.submitted_at:
-                    latest_per_reviewer[review.user.login] = review
-            return max(
-                (
-                    review.submitted_at
-                    for review in latest_per_reviewer.values()
-                    if review.state == "CHANGES_REQUESTED"
-                ),
-                default=None,
-            )
-        except Exception as e:
-            print(f"Error getting reviews for PR #{pr.number}: {e}")
-            return None
+        if all_reviews is None:
+            all_reviews = list(pr.get_reviews())
+        # Bot reviews are advisory; COMMENTED does not change stance.
+        latest_per_reviewer = {}
+        for review in all_reviews:
+            if (
+                not review.user
+                or not review.submitted_at
+                or review.user.type == "Bot"
+                or review.state == "COMMENTED"
+            ):
+                continue
+            current = latest_per_reviewer.get(review.user.login)
+            if current is None or review.submitted_at > current.submitted_at:
+                latest_per_reviewer[review.user.login] = review
+        return max(
+            (
+                review.submitted_at
+                for review in latest_per_reviewer.values()
+                if review.state == "CHANGES_REQUESTED"
+            ),
+            default=None,
+        )
 
     def has_bot_comment(self, pr, comment_type, after_date=None, issue_comments=None):
         """Check if this bot has already posted a comment with the given
--- a/.github/actions/bot-autoassign/stale_pr_bot.py
+++ b/.github/actions/bot-autoassign/stale_pr_bot.py
@@ -13,7 +13,7 @@ class StalePRBot(GitHubBot):
         super().__init__()
         self.DAYS_BEFORE_STALE_WARNING = 7
         self.DAYS_BEFORE_UNASSIGN = 14
-        self.DAYS_BEFORE_CLOSE = 60
+        self.DAYS_BEFORE_FINAL_FOLLOWUP = 60
 
     @staticmethod
     def _commit_activity_date_for_author(commit, pr_author):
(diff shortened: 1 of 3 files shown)


.github/actions/bot-autoassign/stale_pr_bot.py
pandafy, line 167: Add a docstring like other methods.

.github/actions/bot-autoassign/stale_pr_bot.py
```diff
@@ -160,13 +170,25 @@ def get_last_changes_requested(self, pr, all_reviews=None):
         try:
             if all_reviews is None:
                 all_reviews = list(pr.get_reviews())
-            changes_requested_reviews = [
-                r for r in all_reviews if r.state == "CHANGES_REQUESTED"
+            latest_per_reviewer = {}
+            for r in all_reviews:
```
pandafy: Don't use one letter variable names.

.github/actions/bot-autoassign/stale_pr_bot.py
```diff
@@ -160,13 +170,25 @@ def get_last_changes_requested(self, pr, all_reviews=None):
...
+                if r.state == "COMMENTED":
+                    continue
+                current = latest_per_reviewer.get(r.user.login)
+                if current is None or r.submitted_at > current.submitted_at:
+                    latest_per_reviewer[r.user.login] = r
+            blocking = [
+                r
+                for r in latest_per_reviewer.values()
+                if r.state == "CHANGES_REQUESTED"
             ]
```
pandafy: Add an inline comment explaining why are are skipping review.type == "Bot" or review.state == "COMMENTED". 

What is `r.user.login` and why do we need that?

Add an inline comment for the entire loop to explain what it does (shortly) and emphasize on WHY.

.github/actions/bot-autoassign/stale_pr_bot.py
```diff
@@ -32,11 +32,21 @@ def _get_last_author_activity(
             return None
         last_activity = None
         for commit in pr.get_commits():
```
pandafy: Does `pr.get_commits()` return commits in descending order? i.e. the newest commit should be the first in the list to minimize the execution.

.github/actions/bot-autoassign/stale_pr_bot.py
```diff
@@ -32,11 +32,21 @@ def _get_last_author_activity(
...
+                commit_date = max(author_date, committer_date)
+            else:
+                commit_date = author_date or committer_date
+            if not commit_date or commit_date <= after_date:
+                continue
+            author_login = commit.author.login if commit.author else None
+            committer_login = commit.committer.login if commit.committer else None
+            if author_login == pr_author or committer_login == pr_author:
+                if not last_activity or commit_date > last_activity:
+                    last_activity = commit_date
```
pandafy: This logic can be moved to a private helper method. It will make code more readable.

pandafy (changes requested):
@Eeshu-Yadav I think we can make the code more readable which will reduce maintenance efforts in the long run.

.github/actions/bot-autoassign/stale_pr_bot.py
Eeshu-Yadav: Done - extracted to `_commit_activity_date_for_author`.

.github/actions/bot-autoassign/stale_pr_bot.py
Eeshu-Yadav: Commits come oldest-first, not newest-first. Tried optimizing with reverse-iterate but reverted , when rebases mix up author vs committer dates, breaking early can grab the wrong commit.

.github/actions/bot-autoassign/stale_pr_bot.py
Eeshu-Yadav: Done - `r` → `review`.

.github/actions/bot-autoassign/stale_pr_bot.py
Eeshu-Yadav: done

docs/developer/reusable-github-utils.rst
```diff
@@ -69,10 +69,40 @@ OpenWISP repositories. The bot provides the following features:
...
+   from the same reviewer do not block.
+2. **Inactivity** is the time since the more recent of: the PR author's
+   latest commit, issue comment, review comment, or review after the
+   blocking review; or the blocking review's timestamp if the author has
+   not acted since. For commits the date is taken from whichever identity
+   (author or committer) matches the PR author, so a maintainer rebasing
+   the contributor's commits does not reset the clock.
+3. **Maintainer-court skip.** If a maintainer (``OWNER``, ``MEMBER`` or
+   ``COLLABORATOR``) has commented or reviewed after the contributor's
+   last action, the PR is skipped — the ball is in the maintainer's court.
```
nemesifier: This seems wrong. I think it should be the opposite shouldn't it?

If no maintainer has reviewed, then the ball is in the maintainer's court.

Moreover, we should only target reviews, not comments, as we can comment to ask follow up questions but that doesn't count as a review.

nemesifier:
## Approach is correct but wider than the issue describes

The reported bug is that PRs *waiting for review* are being auto-closed. The PR addresses that with three independent changes:

1. `get_last_changes_requested` now ignores `Bot` reviews and uses each reviewer's *latest non-COMMENTED* state — so a `coderabbitai[bot]` "request changes" no longer marks a PR as blocked, and a human who later approved or dismissed their own review no longer counts.
2. `_commit_activity_date_for_author` now considers both `commit.author` and `commit.committer`, so a maintainer rebasing the contributor's commits doesn't reset the inactivity clock; the contributor's force-push, where both identities match, takes the committer date (always later).
3. `close_stale_pr` is removed entirely; the 60-day stage now posts a `final_followup` comment instead of closing.

Each of (1) and (2) on their own solves the reported issue. (3) is a separate scope decision — "we no longer auto-close anything." That's defensible (it eliminates the worst-case mis-action), but the PR title still says "stop … from closing PRs blocked only by bot reviews," which understates the scope. Worth tightening the PR title and description so reviewers understand auto-close is gone, period.

## Stale label gets stuck and never clears once applied

This is the most user-visible regression introduced by removing auto-close. Two paths reach the same dead end:

**Author pushes a fix without commenting.** `mark_pr_stale` adds the `stale` label and unassigns linked issues at day 14. The author pushes a commit on day 16. On the next daily run, `is_waiting_for_maintainer` returns `True` (the new commit is the latest activity), so `process_stale_prs` skips the PR. Nothing in `stale_pr_bot.py` removes the label.

`PRActivityBot.handle_contributor_activity` (`pr_reopen_bot.py:115`) is the code that *would* clear the label and re-assign issues, but its workflow only triggers on `issue_comment.created` and `pull_request_target.reopened` (`bot-autoassign-pr-reopen.yml:4-7`). There's no handler for `pull_request.synchronize`. A contributor who pushes without commenting stays in the `stale` state.

**The bot reaches stage 3 and then has nothing more to do.** Once `bot:stale` and `bot:final_followup` markers exist with `created_at > last_changes_requested`, every future run skips both stages. The `stale` label and the unassigned-issues state persist until a maintainer manually closes the PR or the author comments to trigger `PRActivityBot`. Previously auto-close cleaned this up at 60 days; now nothing does.

Two cheap fixes:

- Have `process_stale_prs` clear the `stale` label and re-assign issues whenever it observes new author commits past `last_changes_requested`, or
- Trigger `PRActivityBot` (or an equivalent) on `pull_request.synchronize` so a push counts the same as a comment.

## A first-time-processed PR over 60 days old gets two comments at once

`test_pr_first_processed_past_60_days_marks_stale_and_followup` enshrines the behavior: with `last_changes_requested` on Feb 1 and "now" on May 10 (≈99 days), stage 1 is skipped (high gate=14), stage 2 fires (`mark_pr_stale`), and stage 3 fires (`send_final_followup`) — all in the same run. The contributor receives two bot comments back-to-back, plus a label change and an issue unassignment.

For PRs the bot tracked continuously this can't happen; the marker check filters past stages. But for a PR the bot missed (workflow disabled, repo migration, schedule outage), it produces a noisy resurfacing. Gating stage 3 on the existence of stage 2's marker would make the follow-up only fire *after* the stale notification was posted in a previous run, mirroring the natural progression.

## `is_waiting_for_maintainer` fail-opens on API errors

```python
except Exception as e:
    print("Error checking maintainer activity" f" for PR #{pr.number}: {e}")
    return False
```

`False` here means "go ahead and process the PR." So a transient PyGithub or rate-limit error makes the bot *more* aggressive on a PR it can't fully assess. Now that auto-close is gone the worst case is a wrong stale-warning or stale-label, which is recoverable, but the policy should still be: when in doubt, skip. Default to `True` on exception.

`get_days_since_activity` (returns `0`) and `get_last_changes_requested` (returns `None`) have the same fail-open shape — both fall through to "process the PR" — though their blast radius is smaller.

## `has_bot_comment` matches any bot's comment containing the marker

```python
if (
    comment.user
    and comment.user.type == "Bot"
    and marker in comment.body
):
```

Any bot whose comment body contains `<!-- bot:stale -->` (quoted excerpt, fork's bot reusing the convention, etc.) satisfies the gate, and the corresponding stage is silently skipped. Realistic risk is low — the markers are unique to this bot — but the fix is one line: also match the comment author against the workflow's bot identity (`github-actions[bot]`). Pre-existing, but the PR adds a third marker (`final_followup`) and increases the surface area.

## `pr.get_commits()` is iterated in full on every check

`_get_last_author_activity` calls `pr.get_commits()` and iterates the entire paginated result. PyGithub paginates lazily, so for a long-running PR with hundreds of commits each daily run pulls every page just to find the latest one whose author or committer matches `pr_author`. Cost scales with PR commit count × stale-eligible PRs × daily runs.

Worse, `process_stale_prs` calls `_get_last_author_activity` twice per PR — once via `get_days_since_activity` and once via `is_waiting_for_maintainer` — and each call re-iterates `get_commits()` from scratch because (unlike `issue_comments`, `review_comments`, and `all_reviews`) it isn't cached at the per-PR scope. Hoist the commit list to `process_stale_prs` and pass it through like the other three.

## `mark_pr_stale` posts the unassign comment before actually unassigning

```python
pr.create_issue_comment("\n".join(unassign_lines))   # claims issues are unassigned
unassigned_count = self.unassign_linked_issues(pr)   # actually unassigns them
```

If the comment-post succeeds but `unassign_linked_issues` fails (rate limit, permission, etc.), the contributor reads "the linked issue(s) have been unassigned from you" while the issue is still assigned to them. Reorder, or word the comment as "we are unassigning…" instead of "have been unassigned." Pre-existing.

## `send_final_followup` opens with bare `Hi @author,` — others have `👋`

`mark_pr_stale` and `send_stale_warning` open with `Hi @{author} 👋,`. `send_final_followup` opens with `Hi @{author},`. Cosmetic, but reads as if the 60-day comment is deliberately colder. Standardize.

## Two-line `print("..." f" ...")` concatenations

Pattern repeats across the file (`print("Error calculating activity" f" for PR #{pr.number}: {e}")`, `print("Error checking maintainer activity" f" for PR #{pr.number}: {e}")`, etc.). Python concatenates adjacent literals at compile time so it's functional, but the result reads like an accidental tuple and the f-string has a leading space that only makes sense alongside its neighbor. Pre-existing, but the new code copies it verbatim — collapse to single f-strings while you're touching the file.

## Test coverage gaps

- No test for the "PR was stale, author pushed, label not removed" scenario described above. Add one as either a regression test (if you fix the behavior) or as an explicit pin of the current behavior so future contributors don't "fix" it accidentally.
- No test for `is_waiting_for_maintainer` when one of the three GitHub API calls inside it raises. Given the fail-open concern above, this branch deserves coverage.

.github/actions/bot-autoassign/stale_pr_bot.py
```diff
@@ -410,67 +401,90 @@ def process_stale_prs(self):
...
+                        # that state now that the contributor has acted.
+                        if self._clear_stale_label(pr):
+                            self._reassign_unassigned_linked_issues(pr)
                         print(
-                            f"PR #{pr.number}: waiting for"
-                            " maintainer review, skipping"
+                            f"PR #{pr.number}: waiting for maintainer review,"
+                            " skipping"
                         )
                         continue
```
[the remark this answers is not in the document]
Eeshu-Yadav: Thanks @coderabbitai - deferring to a follow-up. Daily cron already handles cleanup within ~24h; wiring synchronize needs structural changes (moving recovery helpers, updating PRActivityBot) that warrant their own PR.

nemesifier:
The latest commit (44b2d2a) addresses most of the prior review feedback. The following items remain.

## Doc claim about push-vs-comment is now wrong

`docs/developer/reusable-github-utils.rst:98-100` says:

> An author comment on the PR clears the label and reassigns the issues; a push alone only resets the inactivity timer.

The new `_clear_stale_label` path in `process_stale_prs` means a push *also* clears the label and reassigns issues on the next daily run. The doc reads as if push doesn't clear, which is no longer true. Rework to something like:

> The label clears and issues are reassigned automatically once the bot observes new author activity (next daily run); an author comment also triggers the immediate `PRActivityBot` path.

## Stale label can still get stuck if the blocking review is later approved/dismissed

`stale_pr_bot.py:400-401`:

```python
if not last_changes_requested:
    continue
```

Scenario: PR marked stale at day 14. At day 20 the reviewer approves (or dismisses their own review). On the next daily run, `get_last_changes_requested` returns `None`, the loop short-circuits before reaching the label-unwind block, and the `stale` label persists with linked issues still unassigned.

Cheap fix: hoist the `_clear_stale_label` + `_reassign_unassigned_linked_issues` check above the `if not last_changes_requested` gate, or run it unconditionally for any PR carrying the `stale` label.

## Partial-unassign produces a false comment

`unassign_linked_issues_helper` (`utils.py:105-111`) swallows per-issue exceptions and returns a partial list, so `mark_pr_stale` only sees a successful call even when some issues failed. The comment still reads "the linked issue(s) have been unassigned from you" while one may still be assigned. The prior review suggested rewording to "we are unassigning...", the wording wasn't changed. Low-impact, but the fix is one string edit.

## Two-line `print` concatenations still present

Most occurrences were cleaned up, but a few survive at `stale_pr_bot.py:327-330`, `413-416`, `429-432`. Trivial.

## Behavioral change worth flagging (not a bug)

`is_waiting_for_maintainer` no longer treats maintainer **comments** as maintainer activity, only reviews. The docstring documents this ("Comments don't count") and `test_maintainer_comment_does_not_count_as_review` pins it. PRs where a maintainer asked for changes only through review comments or issue comments will no longer be considered "in maintainer court", they'll proceed to stale-warning. Worth confirming this matches team norms before merging.

## PR title vs. actual scope

The PR title still says "Stop stale-PR bot from closing PRs blocked only by bot reviews," but the change also removes auto-close entirely and revamps the review-stance logic. Please always keep the PR title/description up to date.

nemesifier (approved):

PR #670, opened 2026-05-09
[fix] Avoid Commitizen circular import in changelog bot

### LINKED ISSUE #669: [ci:fix] Prevent circular import in changelog bot Commitizen validation
## Issue Description
- The changelog bot can fail at runtime while validating generated commit messages against the OpenWISP Commitizen rules.
- The failure happens in `generate_changelog.py` when it imports `OpenWispCommitizen` from `openwisp_utils.releaser.commitizen`.
- That module imports `commitizen.cz.base`, which triggers Commitizen plugin discovery.
- Since `openwisp-utils` already registers `OpenWispCommitizen` as a `commitizen.plugin` entry point, Commitizen tries to import the same module again during initialization.
- This leads to a circular import and raises an error similar to:
  - `AttributeError: partially initialized module 'openwisp_utils.releaser.commitizen' has no attribute 'OpenWispCommitizen'`
- As a result, the GitHub Action for changelog generation can fail before posting the generated suggestion.
- see https://github.com/openwisp/openwisp-controller/actions/runs/25575877428/job/75082832879

### Expected behavior
- The changelog bot should be able to reuse the OpenWISP Commitizen validation rules without triggering Commitizen plugin auto-discovery.

### Proposed direction
- Avoid normal importing of `openwisp_utils.releaser.commitizen` inside the changelog action.
- Load only the plugin class and validation logic in a way that does not initialize Commitizen’s plugin registry.

pushpitkamboj:
## Checklist

Closes #669 

## Description of Changes

- Fixes the changelog bot runtime failure caused by circular import during Commitizen validation.
- Updates `get_openwisp_commitizen()` in `.github/actions/bot-changelog-generator/generate_changelog.py` to avoid normal importing of `openwisp_utils.releaser.commitizen`.
- Loads the plugin class without triggering Commitizen plugin auto-discovery, which prevents the module from being imported recursively through the registered `commitizen.plugin` entry point.
- Keeps the change minimal and localized to the changelog action logic.
- Adds a focused regression test to verify that the plugin can be loaded and used without hitting the circular import path.

### DIFF
--- a/.github/actions/bot-changelog-generator/generate_changelog.py
+++ b/.github/actions/bot-changelog-generator/generate_changelog.py
@@ -23,12 +23,15 @@ Environment Variables:
     GEMINI_MODEL: Model to use (default: 'gemini-2.5-flash-lite')
 """
 
+import importlib.util
 import os
 import re
+import runpy
 import secrets
 import subprocess
 import sys
 from html import escape
+from types import ModuleType
 
 from google import genai
 from google.genai import types
@@ -399,10 +402,46 @@ def build_prompt(
 
 
 def get_openwisp_commitizen():
-    """Load the local OpenWISP Commitizen plugin lazily."""
-    from openwisp_utils.releaser.commitizen import OpenWispCommitizen
+    """Load the plugin without triggering Commitizen plugin auto-discovery."""
+    spec = importlib.util.find_spec("openwisp_utils.releaser.commitizen")
+    if spec is None or not spec.origin:
+        raise ImportError("Could not locate openwisp_utils.releaser.commitizen")
 
-    return OpenWispCommitizen(_CommitizenConfig())
+    class _BaseCommitizen:
+        def __init__(self, config):
+            self.config = config
+
+    class _ValidationResult:
+        def __init__(self, is_valid: bool, errors: list[str] | None = None):
+            self.is_valid = is_valid
+            self.errors = errors or []
+
+    fake_commitizen = ModuleType("commitizen")
+    fake_commitizen_cz = ModuleType("commitizen.cz")
+    fake_commitizen_base = ModuleType("commitizen.cz.base")
+    fake_commitizen_base.BaseCommitizen = _BaseCommitizen
+    fake_commitizen_base.ValidationResult = _ValidationResult
+    fake_commitizen.cz = fake_commitizen_cz
+    fake_commitizen_cz.base = fake_commitizen_base
+
+    previous_modules = {
+        name: sys.modules.get(name)
+        for name in ("commitizen", "commitizen.cz", "commitizen.cz.base")
+    }
+
+    try:
+        sys.modules["commitizen"] = fake_commitizen
+        sys.modules["commitizen.cz"] = fake_commitizen_cz
+        sys.modules["commitizen.cz.base"] = fake_commitizen_base
+        plugin_class = runpy.run_path(spec.origin)["OpenWispCommitizen"]
+    finally:
+        for name, previous in previous_modules.items():
+            if previous is None:
+                sys.modules.pop(name, None)
+            else:
+                sys.modules[name] = previous
+
+    return plugin_class(_CommitizenConfig())
 
 
 def get_commit_message_validation_errors(text: str) -> list[str]:
--- a/.github/actions/bot-changelog-generator/test_generate_changelog.py
+++ b/.github/actions/bot-changelog-generator/test_generate_changelog.py
@@ -20,6 +20,7 @@ from generate_changelog import (  # noqa: E402
     get_changelog_validation_errors,
     get_env_or_exit,
     get_linked_issues,
+    get_openwisp_commitizen,
     get_pr_commits,
     get_pr_details,
     get_pr_diff,
@@ -526,6 +527,24 @@ class TestPostGithubComment(unittest.TestCase):
 
 class TestValidateChangelogOutput(unittest.TestCase):
     """Tests for validate_changelog_output function."""
+
+    def test_loads_commitizen_plugin_without_circular_import(self):
+        plugin = get_openwisp_commitizen()
+        self.assertEqual(plugin.__class__.__name__, "OpenWispCommitizen")
+
+        result = plugin.validate_commit_message(
+            commit_msg=(
+                "[feature] Added new functionality #123\n\n"
+                "Adds the new behavior with a user-focused summary.\n\n"
+                "Closes #123"
+            ),
+            pattern=MagicMock(),
+            allow_abort=False,
+            allowed_prefixes=[],
+            max_msg_length=COMMIT_SUBJECT_LIMIT,
+            commit_hash="TEST",
+        )
+        self.assertTrue(result.is_valid)
 
     @patch("generate_changelog.get_openwisp_commitizen")
     def test_valid_feature_tag_rst(self, mock_get_commitizen):
--- a/.github/actions/bot-changelog-generator/test_generate_changelog.py
+++ b/.github/actions/bot-changelog-generator/test_generate_changelog.py
@@ -2,12 +2,14 @@
 
 import os
 import sys
+from types import SimpleNamespace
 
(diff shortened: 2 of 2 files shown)


nemesifier (changes requested):
## Approach is overengineered for the underlying problem

This is the most important issue with the PR. The fix replaces `from openwisp_utils.releaser.commitizen import OpenWispCommitizen` with a 40-line dance that:

1. Builds fake `commitizen`, `commitizen.cz`, `commitizen.cz.base` `ModuleType` stubs.
2. Defines stub `_BaseCommitizen` and `_ValidationResult` classes inside the function body.
3. Temporarily monkey-patches `sys.modules` to point at the stubs.
4. Calls `runpy.run_path(spec.origin)` to execute the plugin file as a fresh module.
5. Restores `sys.modules` in a `finally` block.

A one-line ordering fix produces the same behavior with no fakes, no `runpy`, no `sys.modules` mutation:

```python
import commitizen.cz  # force plugin discovery to complete first
from openwisp_utils.releaser.commitizen import OpenWispCommitizen
```

Verified locally: pre-importing `commitizen.cz` resolves discovery before `openwisp_utils.releaser.commitizen` is on the import stack, so `ep.load()` can fully import and resolve the attribute. After this, the returned class is the *real* `OpenWispCommitizen` subclass of the *real* `commitizen.cz.base.BaseCommitizen`:

```
Class: OpenWispCommitizen
Module: openwisp_utils.releaser.commitizen
isinstance BaseCommitizen: True
valid: True errors: []
```

The PR's approach is correct in outcome but introduces fragility (see below) that the simple ordering fix does not. There is no justification in the description, commit message, or code comments for why the heavier path was chosen.

## Plugin no longer inherits from the real `BaseCommitizen`

With `runpy.run_path` plus the stubbed `commitizen.cz.base`, the returned class's MRO is:

```
[OpenWispCommitizen, _BaseCommitizen, object]
```

instead of `[OpenWispCommitizen, BaseCommitizen, object]` from the real package. Consequences:

* `isinstance(plugin, commitizen.cz.base.BaseCommitizen)` returns `False`.
* Class-level attributes the real `BaseCommitizen` provides (`bump_pattern`, `default_style_config`, `commit_parser`, `template_loader`, `style`, etc.) are missing. Any future call into the plugin that relies on inherited behavior will silently fail or raise `AttributeError`.
* The class's `__module__` is `<run_path>` rather than `openwisp_utils.releaser.commitizen`, which breaks tracebacks, pickling, and tooling that introspects modules.
* The plugin's `validate_commit_message` returns instances of the stub `_ValidationResult` (a regular class), not the real `ValidationResult` (a `NamedTuple`). Tuple unpacking, `_asdict()`, and structural typing checks against `NamedTuple` would all fail.

These are not theoretical: any future enhancement of the plugin or its callers that assumes the real base classes will quietly diverge from the live runtime behavior.

## `_CommitizenConfig` is now silently invalid

`_CommitizenConfig.settings = {}` does not satisfy what real `BaseCommitizen.__init__` does (`self.config.settings.update({"style": ...})`). With the runpy/stub approach the stub `_BaseCommitizen.__init__` doesn't touch `settings`, so the mismatch is masked. If anyone later switches back to the simpler import or accidentally uses the real plugin, the config will need to be updated. There's no comment indicating this dependency.

## `runpy.run_path` re-executes the file on every call

`get_openwisp_commitizen()` is called once per validation in the action's main flow, but tests call it directly. Each call re-reads and re-execs the plugin source file from disk, recreates fake modules, and mutates `sys.modules`. The previous implementation was a cached `from … import` after the first call. Caching is straightforward to add (module-level `_PLUGIN_CLASS = None`), and the omission is not justified.

## Test placement is wrong

The three new tests (`test_loads_commitizen_plugin_without_circular_import`, `test_raises_if_commitizen_plugin_module_cannot_be_located`, `test_restores_sys_modules_if_plugin_loading_fails`) exercise `get_openwisp_commitizen`, not `validate_changelog_output`. They sit inside `class TestValidateChangelogOutput`. They should live in a new `class TestGetOpenwispCommitizen(unittest.TestCase)` for accurate test discovery and reporting.

## `test_restores_sys_modules_if_plugin_loading_fails` is fragile to ordering

The assertion:

```python
self.assertNotIn("commitizen.cz.base", generate_changelog.sys.modules)
```

passes only if no earlier test in the same process imported `commitizen.cz.base` directly. It currently passes because the suite avoids real commitizen imports until `test_loads_commitizen_plugin_without_circular_import` (which uses runpy). If a future test or import order change loads the real `commitizen.cz.base` before this test runs, the assertion will spuriously fail. Either skip this assertion or scope `patch.dict` over `sys.modules` more strictly so the test is independent of process-wide state.

## Test does not verify cleanup actually had something to clean

In `test_restores_sys_modules_if_plugin_loading_fails`, `runpy.run_path` is mocked to raise *before* the function adds the fake `commitizen.cz.base` to `sys.modules` for real (the test mocks runpy after the assignment, so the assignments do happen — but the test never asserts the assignments happened mid-flight). It would be stronger to capture `sys.modules['commitizen']` while inside `runpy.run_path` (e.g., set the side_effect to inspect/mutate then raise) to prove the swap happened and then was reversed.

## `ImportError` message is misleading for namespace packages

```python
if spec is None or not spec.origin:
    raise ImportError("Could not locate openwisp_utils.releaser.commitizen")
```

A spec with `origin is None` happens for namespace packages and similar; the failure mode is "file has no source path", not "module not found". The message conflates two situations. Minor, but worth distinguishing.

## Use of `runpy` is unusual for plugin loading

`runpy.run_path` is intended for executing scripts, not for loading a class out of a packaged module. The conventional API is `importlib.util.module_from_spec` plus `spec.loader.exec_module`, which preserves `__name__`, `__loader__`, and lets the module be cached in `sys.modules` if desired. This is a code-smell that future readers will find disorienting.

## Dead/unused parameter `pattern` (pre-existing, but newly exercised by tests)

The plugin's `validate_commit_message` signature accepts `pattern: re.Pattern[str]`, but the method body never uses it. The PR's new test passes `pattern=MagicMock()` which masks the unused-parameter smell. Not introduced by this PR, but the new test surfaces it. Worth either using `pattern` inside the method or removing it from the signature.

## Documentation drift

The docstring `"""Load the plugin without triggering Commitizen plugin auto-discovery."""` is accurate but doesn't tell the reader why this is needed (the circular import). A one-line `# See issue #669 — circular import via commitizen.plugin entry point` would prevent a future maintainer from "simplifying" the function back into the broken form. The PR description references the issue, but PR descriptions don't survive into the source tree.

## Recommendation

Replace the entire `get_openwisp_commitizen()` body with the ordering fix:

```python
def get_openwisp_commitizen():
    """Load the OpenWISP Commitizen plugin lazily.

    Imports ``commitizen.cz`` first so that the plugin auto-discovery in
    ``commitizen.cz.__init__`` finishes before ``openwisp_utils.releaser.commitizen``
    is added to ``sys.modules``; otherwise the discovery re-imports our module
    while it is still mid-initialization (see issue #669).
    """
    import commitizen.cz  # noqa: F401  -- force plugin discovery first
    from openwisp_utils.releaser.commitizen import OpenWispCommitizen

    return OpenWispCommitizen(_CommitizenConfig())
```

This:

* Fixes the bug.
* Returns the real `OpenWispCommitizen` (correct MRO, correct `__module__`, real `ValidationResult`).
* Eliminates `runpy`, the fake modules, the `sys.modules` mutation, the new `importlib.util` and `runpy` imports, and the stub classes.
* Removes the need for two of the three new tests; the remaining test (`test_loads_commitizen_plugin_without_circular_import`) becomes a clean regression test against the real circular-import scenario.

If the project has reasons to avoid populating the commitizen plugin registry at runtime (e.g., to keep startup time down), the simpler alternative is still preferable but a comment justifying the heavier approach should be added — none currently exists.

nemesifier:
Closed in favour of https://github.com/openwisp/openwisp-utils/pull/671.

PR #674, opened 2026-05-14
[ci] Added retries for changelog generation

pushpitkamboj:
## Checklist

## Description of Changes

- Tightened the changelog bot prompt to more strictly follow OpenWISP commit message and Commitizen validation rules.
- Added retry-based regeneration up to 3 attempts, feeding validation errors back into the prompt to improve the next draft.
- Changed fallback behavior so that if validation still fails on the 3rd attempt, the bot posts the last generated changelog anyway and emits a warning in CI logs.
- Added focused tests covering retry behavior, validation-feedback prompts, and final fallback handling.


.github/actions/bot-changelog-generator/generate_changelog.py
```diff
@@ -370,13 +389,16 @@ def build_prompt(
         "Length: Keep the subject short, but provide enough body detail to help "
```
nemesifier: This new sentence helps, but I think the earlier wording still undermines it because the prompt says the first line should stay within 72 characters "when possible". The validator rejects that case strictly, so I would make this whole section describe hard rules instead of preferences.

I think we can obtain better results with something like this:

```text
Your output will be validated by OpenWISP Commitizen before it is posted.
If any rule below is broken, the output is invalid.

Rules you must follow:

1. Output exactly one plain text git commit message.
2. The first line must start with exactly one of these tags: [feature], [fix], [change].
3. The first line must be 72 characters or less, including the tag and spaces.
4. The first word after the tag must start with a capital letter.
5. The message must contain a body after one blank line.
6. If an issue number appears in the title, the same issue number must appear in the body footer.
7. If an issue number appears in the body footer, the same issue number must appear in the title.
8. Do not use the PR number as an issue number unless it appears in the linked issue data.
9. Do not output Markdown links, ReStructuredText links, GitHub URLs, headings, code fences or explanations.
10. Do not output the GitHub comment intro text.

Issue footer examples that pass validation:

Closes #123
Fixes #123
Resolves #123
Related to #123
```

.github/actions/bot-changelog-generator/generate_changelog.py
```diff
@@ -314,6 +317,18 @@ def build_prompt(
         labels_text = f"\nLabels: {', '.join(safe_labels)}"
     file_name = "CHANGES.md" if changelog_format == "md" else "CHANGES.rst"
     commit_message_rules = build_commit_message_rules_context()
+    validation_feedback = ""
+    if validation_errors:
+        feedback_items = "\n".join(
+            f"- {escape(error, quote=False)}" for error in validation_errors
+        )
+        validation_feedback = (
```
nemesifier, line 325: I would make the retry feedback more direct too. The current wording is understandable, but I think the LLM is more likely to comply if we explicitly say that the previous output failed Commitizen validation and that the next answer must only be the corrected commit message.

Something like this should be clearer:

```text
The previous answer failed OpenWISP Commitizen validation.
Return a corrected commit message only.
Do not explain the fix.
Every validation error below must be fixed:

<validation errors here>
```

nemesifier (changes requested):
Thanks, this is moving in the right direction. The retry loop and passing the validator errors back to Gemini are the right approach, and I agree with keeping this workflow non blocking so it does not distract developers with random CI failures.

I am requesting changes because I think the prompt still makes one strict Commitizen rule sound optional. Tightening that wording should make the first attempt more likely to pass and make the retries more effective.

nemesifier (approved):

PR #692, opened 2026-06-03
[fix:qa] Excluded node_modules and hidden directories from ReStructuredText check

nemesifier:
Fixes the RST QA check so files inside `node_modules/` are skipped reliably.

The previous `--extend-exclude node_modules` approach did not skip deeply nested files, so this uses `find -prune` before calling `docstrfmt`.

The regression test also checks that hidden directories are not included and that the success message is printed once.


nemesifier:
@coderabbitai any outstanding issue? Review latest changes and if no issue is found, resolve.

nemesifier:
@coderabbitai resolve

PR #704, opened 2026-06-22
[ci] Prevent triggering useless CI jobs on forks

asmodehn:
## Checklist

- [X] I have read the [OpenWISP Contributing Guidelines](http://openwisp.io/docs/developer/contributing.html).
- [ ] I have manually tested the changes proposed in this pull request.

I recently noticed the pipelines triggered on my forks. This is the version I am currently checking to not have meaningless CI runs on forks.
Let me know if some changes seem unnecessary.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -17,7 +17,9 @@ permissions:
 
 jobs:
   backport-on-push:
-    if: github.repository == 'openwisp/openwisp-utils' && github.event_name == 'push'
+    if: >
+      github.repository == 'openwisp/openwisp-utils' &&
+      github.event_name == 'push'
     uses: ./.github/workflows/reusable-backport.yml
     with:
       commit_sha: ${{ github.sha }}
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -17,7 +17,7 @@ permissions:
 
 jobs:
   backport-on-push:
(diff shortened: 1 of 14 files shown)


.github/workflows/backport.yml
```diff
@@ -27,6 +27,7 @@ jobs:
 
   backport-on-comment:
     if: >
+      github.repository == 'openwisp/openwisp-utils' &&
```
nemesifier, line 32: @asmodehn what do you think of reformatting all the long ifs as this one to improve readability?

nemesifier (commented):
I think we need to update the docs page:
https://github.com/openwisp/openwisp-utils/blob/master/docs/developer/reusable-github-utils.rst

.github/workflows/backport.yml
asmodehn: Don't we have a yaml formatter that can take care of that ?
I have a yaml checker on my side since i use precommit... but I think I have also seen a yamllint used in some OpenWISP repo... there : https://github.com/openwisp/ansible-openwisp2/blob/master/run-qa-checks#L9

How about setting this up in some form in here ?

asmodehn:
I did a bit of (AI assisted) if guards cleanup. For reference : https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idif

nemesifier (approved):

PR #707, opened 2026-06-24
[feature] Support Python projects using pyproject.toml #706

### LINKED ISSUE #706: [feature] Support Python projects using pyproject.toml

The releaser tool currently detects Python packages by the presence of `setup.py` and reads the version from a `VERSION` tuple in the package's `__init__.py` (or `version.py`). Modern Python projects are increasingly adopting `pyproject.toml` as the single source of truth for metadata, often using it to define project metadata and version via [PEP 621](https://peps.python.org/pep-0621/) (e.g., `[project] version = "x.y.z"`). The releaser does not currently support this setup.

It would be nice to extend the releaser to support Python projects that use `pyproject.toml` for their build configuration and version declaration.

**Describe alternatives you've considered**

This is an open-ended idea meant to start a discussion. Contributors are encouraged to think about:

- How should version recognition work for `pyproject.toml`-based projects? Should the releaser parse the `version` field from `[project]` in `pyproject.toml`, or look for a dynamic version reference?
- How would version bumping work: would the releaser update the version string directly in `pyproject.toml`?
- Would it be possible (and should we) preserve the current pattern used by OpenWISP modules where the canonical version is defined as a `VERSION` tuple in `__init__.py` and then referenced dynamically in `pyproject.toml` (e.g., via `setuptools` dynamic versioning)? If so, how would the releaser detect and bump the version in `__init__.py`?

**Additional context**

The relevant code is in [`openwisp_utils/releaser/config.py`](https://github.com/openwisp/openwisp-utils/blob/master/openwisp_utils/releaser/config.py) (package type detection via `get_package_type_from_setup()`, Python version detection via `_handle_python_version()`) and [`openwisp_utils/releaser/version.py`](https://github.com/openwisp/openwisp-utils/blob/master/openwisp_utils/releaser/version.py) (version bumping via `_bump_python_version()`).

BHARATH0153:
closes #706 

Changes:
- Detect pyproject.toml as a Python package marker (setup.py still takes priority)
- Read version from [project] version field in pyproject.toml (PEP 621)
- Bump version by updating the version string in pyproject.toml
- Graceful fallback when tomllib is not available (Python <3.11)
- Added tests for detection, version reading, version bumping, and priority

## Checklist


BHARATH0153:
@pandafy @nemesifier  had a coverage drop from the pyproject.toml changes. The cause is the conditional import tomllib block — on Python 3.11+ the except ImportError branch is never executed (tomllib is built-in), on 3.10 the import tomllib line is never reached (it fails and falls through to tomli).  can't cover both paths on any single Python version.
so I have used  pragma: no cover on only those version-dependent import branches. Zero business logic is skipped. This is a broadly accepted pattern — openwisp already uses it in admin.py:29. The 5 remaining uncovered lines in the releaser (config.py:112/186/212, version.py:92-93/179) are pre-existing in npm/ansible/openwrt handlers, unrelated to this PR.

BHARATH0153:
@coderabbitai review please

BHARATH0153:
@pandafy please need your perspective check when you have time thanks!

BHARATH0153:
@nemesifier once look into it please thanks!

openwisp_utils/releaser/version.py
```diff
@@ -54,6 +62,28 @@ def _bump_python_version(content, new_version, version_path):
...
+        try:
+            data = tomllib.loads(content)
+        except Exception:
+            pass
+        else:
+            current = data.get("project", {}).get("version", "")
+            if current:
+                old = f'version = "{current}"'
+                if old in content:
+                    return content.replace(old, f'version = "{new_version}"', 1)
```
CodingWithSaksham, line 77: This replaces the first `version = x.y.z` anywhere in the file, not scoped to `[project]`.

openwisp_utils/releaser/config.py
```diff
@@ -203,6 +239,7 @@ def _handle_generic_version(config):
 # Maps package types to their version detection handlers
 PACKAGE_VERSION_HANDLERS = {
     "python": _handle_python_version,
+    "pyproject": _handle_pyproject_toml_version,
```
CodingWithSaksham, line 242: `PACKAGE_VERSION_HANDLERS["pyproject"]` is never reached via `load_config()`
`get_package_type_from_setup()` maps `pyproject.toml` to "python", not "pyproject". So `load_config()` never looks up `PACKAGE_VERSION_HANDLERS["pyproject"]` via the handler dispatch. The "pyproject" package type only appears after `_handle_pyproject_toml_version` sets `config["package_type"] = "pyproject"` inside itself, and this type is subsequently used by `bump_version()` (which is correct). But the "pyproject" entry in `PACKAGE_VERSION_HANDLERS` is dead code for detection purposes. It's harmless but worth clarifying,  either document this intent (the key exists for symmetry/direct testing) or remove it.

openwisp_utils/releaser/tests/test_config.py
CodingWithSaksham, line 1: This test case has been missed I believe:
`setup.py + __init__.py with no VERSION tuple + pyproject.toml with a version`

Please add it to confirm that the fallback to `_handle_pyproject_toml_version` inside `_handle_python_version` actually works end-to-end.

CodingWithSaksham (changes requested):
Hi @BHARATH0153, I have reviewed your PR and the overall design is solid and this is a clean first step.

That said, before a revised PR lands, I would prefer to step back and have a broader discussion on issue #706 first. The reason is that `pyproject.toml` support is not just a releaser feature, it is the opening move of a larger migration that we want to get right. There are multiple cases that can happen like a mixture of `setup.py` and `pyproject.toml` (your PR), fully dropping `setup.py` and migrating to `pyproject.toml` and its PEP styles, or just remaining on `setup.py`.

BHARATH0153:
@CodingWithSaksham let's wait for the @nemesifier @pandafy review

pandafy:
@coderabbitai summary

PR #714, opened 2026-07-04
[change] Make menu item registration idempotent #641

### LINKED ISSUE #641: [change] Make register_menu_group(item) idempotent

When attempting to use `django.test.override_settings(<INSTALLED_APPS>)` in tests, it fails.

```
django.core.exceptions.ImproperlyConfigured: A group/link with config {...} is being registered at position "40",                but another group named "Users & Organizations" is already registered at the same position.
```

The current problem is that the application is reinitialized (ready called a second time), and register_menu_group fails, since it is already there at the specific location.

Ideally, register_menu_group(item) should be idempotent, so calling it multiple times would just ensure it worked. Not crash with an error, when everything is actually in place in the menu.

The simple way to achieve this seems to be skipping the error, if it is the exact same item which is already there, that is being registered again. Just replacing it with the new one (in case the old one is broken for some reason known to the user calling register_menu_group)

**Additional context**
I seem to remember I already encountered this problem in the past, and there was an issue somewhere... Can't find it now though, and I am hitting this problem again on openwisp modules v1.2

CodingWithSaksham:
Re-initializing the Django app registery calls `AppConfig.ready()` again which re-invokes `register_menu_group()` / `register_menu_subitem()` with the same config. Both functions unconditionally raised `ImproperlyConfigured` whenever the target position was already occupied, even when it was the exact same app re-registering the exact same item, making it impossible to survive repeated app registry population.

Both functions now compare the incoming config against the config of the item already registered at that position and log at INFO level if identical config is found Closes

## Checklist

Closes #641 .

## Description of Changes
- Previously, calling `register_menu_group/register_menu_subitem` twice with the same config at the same position always errored, even with app reloads or repeated ready() calls. Now each menu item stores its config, and both functions check for an identical config before raising, if it matches, they log an INFO message and skip instead of erroring. 
- Tests added for the new skip-and-log behavior.


CodingWithSaksham:
@coderabbitai review

nemesifier (changes requested):
As @coderabbitai wrote in its pre-merge checks:

> Update the navigation-menu docs to describe the idempotent identical-config case and keep the error description only for mismatched configs.

asmodehn:
@coderabbitai full review

nemesifier (approved):
Thanks @CodingWithSaksham

PR #731, opened 2026-08-12
[fix] Reload DRF settings after configuring defaults

pandafy:
## Checklist

- [x] I have read the [OpenWISP Contributing Guidelines](https://openwisp.io/docs/dev/developer/contributing.html) and [OpenWISP Anti AI Spam Policy](https://openwisp.io/docs/dev/general/code-of-conduct.html#anti-ai-spam-policy).
- N/A I have updated the documentation.

N/A

## Description of Changes

Bug:
`ApiAppConfig.configure_rest_framework_defaults` updated Django's `settings.REST_FRAMEWORK`, but DRF's `api_settings` had already loaded its own copy of the configuration. As a result, changes made by apps initialized after `rest_framework` were not reflected in the settings actually used by DRF.

Fix:
Reload `rest_framework.settings.api_settings` after updating `settings.REST_FRAMEWORK` so DRF uses the latest configuration.

### DIFF
--- a/openwisp_utils/api/apps.py
+++ b/openwisp_utils/api/apps.py
@@ -14,6 +14,7 @@ class ApiAppConfig(AppConfig):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        # Configure defaults before Django imports models, which may cache DRF settings.
         if self.api_enabled:
             self.configure_rest_framework_defaults()
 
@@ -43,9 +44,8 @@ class ApiAppConfig(AppConfig):
             # otherwise just set it as default value
             current_settings.setdefault(key, value)
 
-        # explicitly set it in settings.py
+        # Set merged defaults before DRF reads and caches them.
         setattr(settings, "REST_FRAMEWORK", current_settings)
-
         from rest_framework.settings import api_settings
 
         api_settings.reload()
--- a/openwisp_utils/api/apps.py
+++ b/openwisp_utils/api/apps.py
@@ -1,6 +1,7 @@
 from django.apps import AppConfig
 from django.conf import settings
 from rest_framework.settings import api_settings
+from rest_framework.views import APIView
 
 from ..utils import deep_merge_dicts, default_or_test
 
@@ -46,3 +47,4 @@ class ApiAppConfig(AppConfig):
         # explicitly set it in settings.py
         setattr(settings, "REST_FRAMEWORK", current_settings)
         api_settings.reload()
+        APIView.throttle_classes = api_settings.DEFAULT_THROTTLE_CLASSES
--- a/openwisp_utils/api/apps.py
+++ b/openwisp_utils/api/apps.py
@@ -1,7 +1,5 @@
 from django.apps import AppConfig
 from django.conf import settings
-from rest_framework.settings import api_settings
-from rest_framework.views import APIView
 
 from ..utils import deep_merge_dicts, default_or_test
 
@@ -14,7 +12,8 @@ DEFAULT_REST_FRAMEWORK_SETTINGS = {
 class ApiAppConfig(AppConfig):
     API_ENABLED = False
 
-    def ready(self, *args, **kwargs):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
         if self.api_enabled:
             self.configure_rest_framework_defaults()
 
@@ -46,5 +45,7 @@ class ApiAppConfig(AppConfig):
 
         # explicitly set it in settings.py
         setattr(settings, "REST_FRAMEWORK", current_settings)
+
+        from rest_framework.settings import api_settings
+
         api_settings.reload()
-        APIView.throttle_classes = api_settings.DEFAULT_THROTTLE_CLASSES
--- a/openwisp_utils/api/apps.py
+++ b/openwisp_utils/api/apps.py
@@ -1,5 +1,6 @@
 from django.apps import AppConfig
 from django.conf import settings
+from rest_framework.settings import api_settings
 
 from ..utils import deep_merge_dicts, default_or_test
 
@@ -44,3 +45,4 @@ class ApiAppConfig(AppConfig):
 
         # explicitly set it in settings.py
         setattr(settings, "REST_FRAMEWORK", current_settings)
+        api_settings.reload()
--- a/tests/api_app_config/first/models.py
+++ b/tests/api_app_config/first/models.py
@@ -1,0 +1,2 @@
+from rest_framework.throttling import SimpleRateThrottle  # noqa: F401
+from rest_framework.views import APIView  # noqa: F401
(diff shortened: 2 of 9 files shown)


openwisp_utils/api/apps.py
```diff
@@ -1,5 +1,6 @@
 from django.apps import AppConfig
 from django.conf import settings
+from rest_framework.settings import api_settings
```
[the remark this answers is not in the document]
pandafy: Isn't `rest_framework` a hard dependency for any module which wants to implement REST APIs? @coderabbitai

openwisp_utils/api/apps.py
```diff
@@ -44,3 +46,5 @@ def configure_rest_framework_defaults(self):
 
         # explicitly set it in settings.py
         setattr(settings, "REST_FRAMEWORK", current_settings)
+        api_settings.reload()
+        APIView.throttle_classes = api_settings.DEFAULT_THROTTLE_CLASSES
```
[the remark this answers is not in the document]
pandafy: @coderabbitai rest_framework.settings.api_settings is accessed by a lot of objects in DRF. We cannot update all the objects directly here.

openwisp_utils/api/apps.py
```diff
@@ -1,5 +1,7 @@
 from django.apps import AppConfig
 from django.conf import settings
+from rest_framework.settings import api_settings
+from rest_framework.views import APIView
```
nemesifier: I think this workaround introduces a larger initialization bug than the one this PR is fixing. Importing `APIView` while Django imports app configs resolves all policy settings before any `ready()` method merges `REST_FRAMEWORK_SETTINGS`, and line 50 refreshes only `throttle_classes`. I reproduced `api_settings.DEFAULT_PERMISSION_CLASSES == [IsAuthenticated]` while `APIView.permission_classes == [AllowAny]`. I also reproduced `ScopedRateThrottle.get_rate()` raising `ImproperlyConfigured` because `SimpleRateThrottle.THROTTLE_RATES` was cached before the app-added `test` scope.

Please remove these module-level DRF imports and the `APIView.throttle_classes` assignment. Import only `api_settings` inside `configure_rest_framework_defaults()` after the merge, then call `api_settings.reload()`. This smaller change lets DRF policy classes load with the merged settings, makes the `test` scope resolve to `10/minute`, and preserves the optional `rest` extra. Adding more direct cache assignments would remain incomplete.

tests/test_project/tests/test_api.py
```diff
@@ -86,6 +101,24 @@ def test_rest_framework_settings_override(self):
...
+    @override_settings(REST_FRAMEWORK={"DEFAULT_THROTTLE_RATES": {"anon": "20/hour"}})
+    def test_rest_framework_settings_update_existing_api_views(self):
+        self.assertEqual(APIView.throttle_classes, [CustomScopedRateThrottle])
+        apps.get_app_config("test_project").configure_rest_framework_defaults()
+        self.assertEqual(
+            settings.REST_FRAMEWORK["DEFAULT_THROTTLE_CLASSES"],
+            ["rest_framework.throttling.ScopedRateThrottle"],
+        )
+        self.assertEqual(api_settings.DEFAULT_THROTTLE_CLASSES, [ScopedRateThrottle])
+        self.assertEqual(APIView.throttle_classes, [ScopedRateThrottle])
```
nemesifier: This test mutates `APIView.throttle_classes` globally and `override_settings` does not restore it. I reproduced that it starts with `CustomScopedRateThrottle` and leaves `ScopedRateThrottle` behind after the decorator exits.

Please remove this workaround-specific test with the `APIView` assignment. The settings test should instead instantiate a `ScopedRateThrottle` for the app-added `test` scope after `configure_rest_framework_defaults()` and verify its rate is `10/minute`. Merely checking `api_settings` did not catch the runtime failure.

nemesifier (changes requested):
The settings reload is the right direction, but the APIView workaround must be removed. I left inline comments with the smaller fix and the missing behavioral regression coverage.

nemesifier:
@pandafy I reworked this so the merge happens during `ApiAppConfig` construction, before Django imports models and any DRF policy classes. This avoids trying to repair individual DRF caches after `ready()`, which was incomplete: `APIView` snapshots more than throttles and `SimpleRateThrottle` separately snapshots rates.

The new startup fixture imports `APIView` and `SimpleRateThrottle` from a model module, then verifies project precedence, multiple app contributions, permissions, and scoped-rate resolution. It also verifies that an API-disabled config can start without DRF. The direct `APIView` mutation and its order-dependent test are removed.

I kept `api_settings.reload()` for explicit later calls to `configure_rest_framework_defaults()`, but it is no longer relied on to repair classes created during normal startup.

CodeRabbit's only new valid finding was a missing subprocess timeout. I added it in `47e8626`. Local QA, focused coverage, and the full suite pass.

openwisp_utils/api/apps.py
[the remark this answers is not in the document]
nemesifier: Fixed in 3f13752. DRF is now imported only after the API defaults are merged, and the disabled-config startup regression test verifies that this module remains importable without djangorestframework.

openwisp_utils/api/apps.py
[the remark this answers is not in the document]
nemesifier: This is fixed without patching the cache directly. Enabled ApiAppConfig instances now merge defaults during phase 1, before Django imports models and DRF imports SimpleRateThrottle. The clean-startup regression fixture imports SimpleRateThrottle from a model module and verifies the merged rates and scoped lookup.

nemesifier:
@coderabbitai resolve

nemesifier (dismissed):
All review feedback is addressed. The latest CI run and CodeRabbit review are green.

nemesifier:
@coderabbitai review

nemesifier:
@coderabbitai resolve

tests/api_app_config/first/models.py
pandafy, line 1: Any specific reason to keep these imports in models.py instead of views.py?

tests/test_project/tests/test_api.py
```diff
@@ -209,6 +233,84 @@ def test_model_validation_create_and_update(self):
             self.assertEqual(shelf.locked, True)
 
 
+class TestApiAppConfig(TestCase):
+    def test_defaults_are_available_before_models_are_imported(self):
+        code = """
```
pandafy, line 257: A comment / docstring about why we need to implement these test in this particular way would be helpful.

pandafy (commented):
Thanks for picking up my work and improving it @nemesifier!

tests/api_app_config/first/models.py
nemesifier: Kept intentionally. This fixture imports DRF while Django imports models so the startup test verifies that API defaults are configured before DRF creates its class-level setting caches.

tests/test_project/tests/test_api.py
nemesifier: Added a docstring explaining that this must run in a clean subprocess, before Django initialization and DRF settings caches exist.

nemesifier:
@coderabbitai review

nemesifier (approved):

